PR #27, opened 2015-03-24
Removed 3rd party telnet and replace with nodejs net

RonaldZ:
The 3rd party telnet hasn't updated in years so removed it and replaced it with node's net utility.

Runs nearly the same and the change to the code is minimal.

Also removed the require for express, not sure why it was there but the rest of the code had no reference to it.


shawncplus:
It was there because, at the time, windows telnet could not connect to the game with the nodejs builtin. Is this still the case?

RonaldZ:
Would be odd if there is a problem. Let me get a vm spun up so I can test it and I'll get back to you on that.

shawncplus:
If I recall correctly express was originally there because https://github.com/shawncplus/ranvier-webbuilder existed inside the core but I split it out later after deciding to implement a plugin system

RonaldZ:
Gotcha, so definitely no reason for it to be in there then.

Waiting for an iso to download so I can begin the testing for windows. Will be doing Windows 7 since that seems to be the most popular.

RonaldZ:
Took awhile, but seems you can't even run nodejs on Windows 10 without it blowing up. So not going to worry about that.

As for Windows 7, went without a hitch. Ran the server and connected to it locally without any errors or hang ups.

hath995:
Honestly, I wouldn't remove the telnet server code. The telnet specification hasn't changed in years so having a library which hasn't updated lately is fine. Secondly, net is just a tcp connection. It doesn't interpret the data. The Telnet specification includes a fair amount of control sequences which aren't part of a tcp connection. Assuming the existing libraries handles those then removing would mean that you would now need to interpret those. I imagine that old telnet clients which send out those control sequences will either error out or will go into funky modes because acknowledgement sequences weren't received. 

I imagine what will happen if these aren't handled is that you will get user input with a bevy of invisible/non-ascii characters before and after input which you will end up needing to filter out.

shawncplus:
Yeah I'm slightly apprehensive as well. I remember running into weird issues like this one in particular https://github.com/shawncplus/ranviermud/blob/master/src/events.js#L78

RonaldZ:
Odd, I thought what hath is talking about was happening already because I saw that same line.

PR #45, opened 2015-10-28
Deal with errors while trying to start the server

marado:
Specific error info in case of EADDRINUSE or EACCES.


shawncplus:
Good addition

PR #48, opened 2015-11-04
Help

seanohue:
**New command:**

`help` 

Typing in just `help` or `help help` will tell you how help works.
Typing in any other command will explain how that command works.

There is also error handling in case someone adds a custom command without adding documentation, so that the server will not crash. There is a separate message if the argument typed by the player is not actually a command. I'd imagine this command is easily extensible to include strings beyond just the usual commands, but it'd be only a bit of work (in case MUD creators want to include help files on the game's lore, areas, a helpfile for newbs, etc.) -- this could be a future feature. Or something for MUD creators to add, so that this engine stays fairly lightweight.

Incidentally, this PR also fixes a typo in one of the comments in command_util.js


seanohue:
Well, I accidentally merged the master branch of my fork into this, so I'm closing this PR until I fix that.

PR #53, opened 2015-11-08
Replace localize with jall

marado:
localize library is no longer maintained, jall is a retro-compatible fork that is maintained.


seanohue:
Nice catch.

PR #54, opened 2015-11-08
Describe character

seanohue:
Players can now use `describe [description]` to change their character's description. Using `desc` by itself will show the current description.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -1,6 +1,13 @@
 # RanvierMUD
 NodeJS based MUD engine with full localization support
 
+## Contributing to this fork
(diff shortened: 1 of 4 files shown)


seanohue:
Still working on this since it seems the description does not save if the server is reset.

seanohue:
I fixed the issue by having the description be part of the player's attributes.

Xarcell:
I'm not sure how this description works because I haven't tried it yet, but does it include wearables? For example, if I'm wearing a "black leather cloak", then that should be appended to the description. Does it do this?

seanohue:
@Xarcell It does not. This is the written description. I think the look command would have to be improved further to show the character's name, description, and their wearables, rather than appending it to this description, which is essentially flavor text.

seanohue:
@shawncplus This one will cause a merge conflict with one of the previous PRs in player.js

Just keep the line `	self.getDescription  = function () { return self.attributes.description; }; `

and it should work like gravy.

PR #57, opened 2015-11-10
Emote

seanohue:
This command lets players perform an action by typing "emote (action)". This will be displayed in the third person to everyone in the room. For example, `emote jumps with joy` would display `Bob jumps with joy` to all players in the same room as Bob.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -1,6 +1,13 @@
 # RanvierMUD
 NodeJS based MUD engine with full localization support
 
+## Contributing to this fork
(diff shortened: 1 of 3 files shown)


Xarcell:
Are the actions predefined? Or can you put anything. Being able to put anything will cause RP issues. I could write something like: emote sets your cloak on fire and you burn to death, a slow agonizing and painful death. Even though the other person is just fine. I see this being abused if you can put anything...

seanohue:
You can put anything. This is a fairly common feature on other MUDs and people abusing it like that can happen, but would break the unspoken rule of not being a wang or trying to ruin others' fun.

seanohue:
@shawncplus This has probably reached the level of outright spamming, but this one plays well with the previous PRs.

PR #59, opened 2015-12-01
Give command

seanohue:
Players can now give items to other players! There is error handling (maybe too much of it, but better safe than sorry) however this may need to be further smoke tested.

Enjoy.


Xarcell:
Thanks for this!

chrisciampoli:
Awesome thanks!

PR #61, opened 2016-02-01
Get all

seanohue:
Now, typing "get all" will allow you to pick up all of the items in the room as long as you are not in combat, and have room in your inventory.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -1,6 +1,13 @@
 # RanvierMUD
 NodeJS based MUD engine with full localization support
 
+## Contributing to this fork
+This fork aims to eventually turn the lightweight MUD engine Ranvier, developed by shawncplus, into a full-fledged game. The setting is, currently, a post-apocalyptic steampunk port town. That is probably a cliche, but I don't care. Right now, the master branch is working and has a fair amount of new content and features. However, I have a better understanding of the codebase now so it is in flux.
+
+There is also a development branch where I am working on rewriting some of the features to submit to the upstream project. At some point, the dev branch will be updated to include a lot more content. For the time being, it's bare since I am rewriting some features I'd already added.
+
+Please submit any PRs that are related to making this into a game to this project, preferably the master branch. Please submit any features based on enhancing the MUD engine itself to the shawncplus project.
+
 ## Features
 * Full localization for any strings displayed to the player after they log in. This isn't a common feature in MUDs so I'll explain it. Player A and Player B log in, play the same game and see a different language. Same items, npcs, rooms, etc, different language.
 * Scripting support: It's in Javascript! No need for a shitty DSL. The codebase is javascript, the scripting is javascript.
@@ -24,6 +31,8 @@ NodeJS based MUD engine with full localization support
     cd ranviermud
     npm install
     sudo ./ranvier -v --save=10 --respawn=10
+
+Connect to the server using `telnet 23`
     
 
 ## Documentation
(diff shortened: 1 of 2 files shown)


Xarcell:
Thanks for this.

seanohue:
@Xarcell Thanks. Any suggestions on what to tackle next?

Xarcell:
It's hard to say. There are things I'd like to see, but that doesn't mean everyone would want it. Here are a few I can think of off the top of my head: 

1.| Optional 3 different room descriptions for different times of the day, such as dawn/dusk, day, and night.

2.| Optional 2 additional different room descriptions based on racial attributes, some see in the dark and are blinded by light, while others are blind in the dark, but see in the light. If combined with number one, it would total 5 room descriptions per room(optional of course, not every room needs additional descriptions, such as the inside of a tavern.

3.| Throwing = the ability to throw items for damage and recover some those items, but not always.

4.| Item Use Breakage = either through complex formula or just set number of uses.

5.| Guild = accept or deny other players into the guild. Guild houses restrict access to members only.

seanohue:
Nice ideas! I am also trying to work on shops but am unable to fathom how I'd set it up in YML, or if that'd even be the place to start. And related to point #5, I also want to add admin/wizard abilities to be able to make changes in game. Adding guild leader/member permissions would be similar. Another related story could be locked doors -- you can only enter with a certain item in your inventory. 

Thanks for the ideas, @Xarcell

shawncplus:
@seanohue my original goal for in game changes was the web editor here https://github.com/shawncplus/ranvier-webbuilder (Which is equally wicked out of date). I really hate in-game OLEs they're always clunky.

Regarding shops: off the top of my head would just be something like

NPC YML

    -
      vnum: 7
      room: 4
      short_description:
        en: 'A shopkeeper stands at the counter'
      shop: 1
      behaviors: "shopkeeper"

object YML

    - &longsword
        vnum: 1
        short_description:
            en: "a longsword"
            es: "A rapier"
        keywords:
            en: [ "longsword", "sword", "long" ]
            es: [ "rapier", "sword" ]

shop YML (would need a new shops/ dir and data load)

shop/1.yml

    - &newbieshop
      vnum: 1
      pricing_calc: " script to modify price based on game state
        reputation: guardians " imaginary pricing script that takes a faction and gives a % discount based on standing
        quest: 12 " discount for having completed given quest
        class: warrior " discount for given class
      purchase: true " whether players can sell items here
      items:
        -
          item: 1
          price: 100
          inventory: 3 " count in the shop before it runs out, restock on respawn
          multisell: true " alternative to inventory, stock never depletes

seanohue:
@shawncplus I never thought of having both a shop entity and a shopkeeper entity, but that makes perfect sense like that. I also like the various flags and idea of having some shops sell items for more than others.  The shopkeeper behavior script and the buy/sell/list commands would have to all be programmed to play well together, I assume? I still don't fully understand how the items and npcs work and how the MUD grabs all of this information from the YML, but this is a good starting point, thanks.

As an aside, this PR doesn't break any of the previous ones I've submitted.

seanohue:
@shawncplus Also might hack away on a tool for making in-game changes, but probably not that one since I have zero experience with ExtJS. Maybe Angular or something similar.

shawncplus:
@seanohue yeah, please don't bother with that old crap :yum: With that said you can probably copy/paste all of the API stuff that the plugin uses and just rebuild the frontend

seanohue:
@shawncplus Gotcha. Related to that -- is there a copyover function??

shawncplus:
@seanohue there's a really terrible hotboot command that's usable from the server exe while it's running but no in-game command to execute it

PR #62, opened 2016-02-02
Tell command

seanohue:
Noticed there wasn't a "tell" command, figured this could be useful to some.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -1,6 +1,13 @@
 # RanvierMUD
 NodeJS based MUD engine with full localization support
 
+## Contributing to this fork
+This fork aims to eventually turn the lightweight MUD engine Ranvier, developed by shawncplus, into a full-fledged game. The setting is, currently, a post-apocalyptic steampunk port town. That is probably a cliche, but I don't care. Right now, the master branch is working and has a fair amount of new content and features. However, I have a better understanding of the codebase now so it is in flux.
+
+There is also a development branch where I am working on rewriting some of the features to submit to the upstream project. At some point, the dev branch will be updated to include a lot more content. For the time being, it's bare since I am rewriting some features I'd already added.
+
+Please submit any PRs that are related to making this into a game to this project, preferably the master branch. Please submit any features based on enhancing the MUD engine itself to the shawncplus project.
+
 ## Features
 * Full localization for any strings displayed to the player after they log in. This isn't a common feature in MUDs so I'll explain it. Player A and Player B log in, play the same game and see a different language. Same items, npcs, rooms, etc, different language.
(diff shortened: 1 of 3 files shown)


Xarcell:
There are alot of commands I'd like to see. I can post a few tomorrow if your interested?

Xarcell:
One thing I noticed by looking at the code is that the audience doesn't see anything. For example: Sean tells Xarcell blah blah blah. I noticed the same with the give command. Shouldn't the audience(people in the room) also see what Sean is giving Xarcell?

shawncplus:
@Xarcell tell being silent makes sense, that's the purpose of a tell but I agree with give being "loud". Though it brings up the point of having configurable spam filters so players can mute loud commands not targeted at or by them.

seanohue:
@Xarcell Basically, what @shawncplus said -- though, I'll add something for 3rd parties for the 'give' command. Tell is more for one on one OOC communication. They don't even need to be in the same room. Whisper is for in character conversation, and the third party can see that something is being whispered.

Xarcell:
@seanohue ah, I didn't realize tell was a OOC communication.

seanohue:
@Xarcell I've usually seen it used as an inter-player messaging system, like a private message. I guess whether it's OOC or not depends on the MUD, but it's a way to communicate with anyone online regardless of if they are in the same room or not.

Xarcell:
I'm glad you added it. Only thing I worry about is the role playing aspect of it though. In Gemstone III, you needed an crystal amulet that you had to rub it to activate for it 60 seconds in order to communicate private messages or default sales channel. If you sent someone a PM, but your amulet wasn't activated, it would glow as a indicator you had a PM. Then you'd rub it to see you PM. Or something like that, it's been awhile since I played.

shawncplus:
This is just my personal opinion on it since MUDders have different preferences but I prefer function over form and a lot of MUDs that are RP heavy tend to go the other way. There's a reason I made the skill system instantly grant skills at level up with no percentage system or practices, because it's F'ing annoying and gets in the way of the actually fun part of the game. At the same time one of the reasons it's so flexible is because I know, as I said, every MUDder has a different preference.

So as a preference I prefer to have zero-effort OOC utilities (tell/help/newbie/chat) instead of trying to shoehorning RP into it. Imagine if help files required you to sacrifice an item to a demon of knowledge or something, yeah it'd be awesome for RP but _horrible_ for actual usability.

seanohue:
@shawncplus exactly, I figured this was the most 'lightweight' way of doing it, since that's the aim of the project, and the tell command could easily be removed/edited now that it's there in skeletal form.

seanohue:
@xarcell Also, I'd be interested in any ideas you have.

seanohue:
Closed because it's just duplicating the tell channel.

PR #89, opened 2017-01-14
Basic api impl

seanohue:
Addresses parts of #94 

Make a GET request against `hostname:9000/api/[whatever]` to get information.

Right now, you can whitelist by IP address in ranvier.json if you so desire. More authentication is to be added later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ coverage/
 *.log
 *.debug
 gulpTODO.md
+.ranvierSecret
--- a/src/WebInterface.js
+++ b/src/WebInterface.js
@@ -60,28 +60,6 @@ class WebInterface {
     router.get('/help',    this.getResponseData(HelpManager, 'helps'));
   }
 
-  authenticate(req, res) {
-    const accountName = req.body.name;
-    const account = this.state.AccountManager.findByName(accountName);
-    
-    if (!account) {
-      util.log(`[WEB] ${accountName} Auth Failed: Account not found.`);
-      return res.json({ success: false, message: `${accountName} was not found.`});
-    }
-
-    if (!bcrypt.compareSync(req.body.password, account.password)) {
-      util.log(`[WEB] ${accountName} Auth Failed: Incorrect password.`);
-      return res.json({ success: false, message: `${accountName} password incorrect.`});
-    }
-
-    let secret    = this.getSecret();
-    let expiresIn = "24h";
-    const token   = jwt.sign(account, secret, { expiresIn });
-
-    util.log(`[WEB] ${accountName} Auth Success!`);
-    return res.json({ success: true, token, message: `${accountName} authenticated. Token good for ${expiresIn}.`});
-  }
-
   getResponseData(manager, name) {
     return (req, res) => {
       const response = this.parseEntitiesIntoResponse(manager, name);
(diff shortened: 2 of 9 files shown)


seanohue:
screen shot 2017-01-14 at 4 31 50 pm

src/WebInterface.js
```diff
@@ -0,0 +1,105 @@
...
+    router.post('/authenticate', (req, res) => {
+      const accountName = req.body.name;
+      const account = this.state.AccountManager.findByName(accountName);
+      
+      if (!account) {
+        util.log(`[WEB] ${accountName} Auth Failed: Account not found.`);
+        return res.json({ success: false, message: `${accountName} was not found.`});
+      }
+
+      // Both passwords should be hashed... right?
```
shawncplus: The password they send would be plaintext, we would have to hash it and compare

.gitignore
```diff
@@ -10,3 +10,4 @@ coverage/
 *.log
 *.debug
 gulpTODO.md
+.ranvierSecret
```
shawncplus: Want to add to the README setup section info on creating the JWT secret?

src/WebInterface.js
seanohue: Do we really want to send passwords in plaintext? I would think it would be hashed on the client side and sent over as a hash, using the same algorithm.

.gitignore
seanohue: Sure. In this case I just used a random word but I'd assume they'd want some kind of randomly generated text in a file that could be shared between server and client. I'll write a thing about it later today.

src/WebInterface.js
shawncplus: Passwords would be sent by the client in plaintext preferably over HTTPS just like any website. It would be terrible to have the clientside hash the passwords.

src/WebInterface.js
```diff
@@ -14,92 +15,113 @@ const Config = require('./Config');
...
+    return this.secret;
+  }
 
+  setUpGetRoutes() {
+    const message = { message: "RanvierMUD API Endpoints: /npcs /players /items /rooms /help" };
+    const { MobFactory, PlayerManager, ItemManager, RoomManager, HelpManager } = this.state;
 
-    app.use('/api', router);
+    // Routes for the API's GET response.
+    router.get('/',        (req, res) => res.json(message));
```
shawncplus: This line here is pretty non-standard unless you're doing HATEOAS. This should either 404 because there's nothing to do or if you want to keep this message make the response code 405 to indicate that it's not supported

src/WebInterface.js
```diff
@@ -14,92 +15,113 @@ const Config = require('./Config');
...
-    router.post('/authenticate', (req, res) => {
-      const accountName = req.body.name;
-      const account = this.state.AccountManager.findByName(accountName);
-      
-      if (!account) {
-        util.log(`[WEB] ${accountName} Auth Failed: Account not found.`);
-        return res.json({ success: false, message: `${accountName} was not found.`});
-      }
+    // Authentication route.
+    router.post('/authenticate', (req, res) => this.authenticate(req, res));
```
shawncplus: It looks like we have the authenticate endpoint to get the jwt but then we're not actually verifying it on subsequent requests?

seanohue: Right. So here's where I was looking for feedback -- I am wondering if only POST routes (to make changes) should be authenticated, or if all of them should, or what. Maybe it could be configurable in `ranvier.json`. 

Though, it probably makes sense to just protect all of the routes.

Also, I realized we are currently doing nothing with the "Secret" besides using it to make a token.

src/WebInterface.js
seanohue: Good point, I will probably just remove it.

PR #97, opened 2017-01-30
Allow multiplay

### LINKED ISSUE #76: Don't log off characters from a player when he logs in
This would allow for a configuration setting letting the MUD admins decide whether or not to allow multiple characters from the same user logged in at the same time.
It would also allow session takeovers (log in into a character that is already logged in, and the new session takes over the old one).

seanohue:
Looks like there are some merge conflicts that I will try to fix on my end before pushing this up again.

Fixed #76 and made multiplay and max characters per account configurable.

Behavior for multiplay-allowed:
* Boots the old character and logs in a new session if you try to log in as a character that is already online.
* Else, lets you sign into the new and old character at once.

Behavior for multiplay-banned:
* Lets you sign into your account without logging off an old player, but as soon as you select the character it will kick your old char.

### DIFF
--- a/bundles/core-input-events/input-events/choose-character.js
+++ b/bundles/core-input-events/input-events/choose-character.js
@@ -40,7 +40,10 @@ module.exports = (srcPath) => {
       if (canAddCharacter) {
         options.push({
           display: 'Create New Character',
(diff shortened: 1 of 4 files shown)


seanohue:
Conflicts should be resolved as of now.

PR #104, opened 2017-02-07
Attributes Class & Configurable Attributes/Defaults.

seanohue:
Default attributes for all characters are now configurable in `ranvier.json`. There is now a generic Attributes class, which ends up being a Map with some proxying. The Attribute methods are then wrapped by methods of the Character.

It should be relatively easy to also implement attributes for Items, using the same base class.

This update also includes changes to any spot in the code that directly accessed properties of Attributes, instead using the methods to access and set values. This better fits the way Effects are currently implemented, too.

### DIFF
--- a/bundles/core-combat/player-events.js
+++ b/bundles/core-combat/player-events.js
@@ -17,12 +17,12 @@ module.exports = (srcPath) => {
         // could potentially wind up in a situation where the player performed
         // a mid-round attack that killed the target, then the next round the
         // target kills the player. So let's not let that happen.
-        if (this.attributes.health <= 0) {
+        if (this.getAttribute('health') <= 0) {
           this.combatants.forEach(combatant => {
             this.removeCombatant(combatant);
             combatant.removeCombatant(this);
             if (!combatant.isInCombat()) {
-              combatant.attributes.health = combatant.attributes.maxHealth;
+              combatant.setAttribute('health', combatant.getRawAttribute('maxHealth'));
             }
           }, this);
 
@@ -49,7 +49,7 @@ module.exports = (srcPath) => {
         let hadActions = false;
         for (const target of this.combatants) {
           // player actions
-          if (target.attributes.health <= 0) {
+          if (target.getAttribute('health') <= 0) {
             this.removeCombatant(target);
             target.removeCombatant(this);
 
@@ -71,10 +71,11 @@ module.exports = (srcPath) => {
 
           if (this.combatData.lag <= 0) {
             hadActions = true;
-            const damage = Math.min(playerDamage, target.attributes.health);
+            const damage = Math.min(playerDamage, target.getAttribute('health'));
 
             target.emit('hit', this, damage);
-            target.attributes.health -= damage;
+            const startingHealth = target.getRawAttribute('health');
+            target.setAttribute('health', startingHealth - damage);
             Broadcast.sayAt(this, `You strike <bold>${target.name}</bold> for <bold>${damage}</bold> damage`);
 
             this.combatData.lag = playerSpeed * 1000;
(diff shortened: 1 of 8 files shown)


bundles/core-combat/player-events.js
```diff
@@ -17,12 +17,12 @@ module.exports = (srcPath) => {
...
         // a mid-round attack that killed the target, then the next round the
         // target kills the player. So let's not let that happen.
-        if (this.attributes.health <= 0) {
+        if (this.getAttribute('health') <= 0) {
           this.combatants.forEach(combatant => {
             this.removeCombatant(combatant);
             combatant.removeCombatant(this);
             if (!combatant.isInCombat()) {
-              combatant.attributes.health = combatant.attributes.maxHealth;
+              combatant.setAttribute('health', combatant.getRawAttribute('maxHealth'));
```
shawncplus: This should probably be `getAttribute` instead of `getRawAttribute` in, for example, there was an effect on the Npc that increased or lowered its max health

bundles/core-combat/player-events.js
```diff
@@ -111,7 +115,7 @@ module.exports = (srcPath) => {
           this.combatData = {};
 
           // TODO: There is no regen at the moment so if they won just reset their health
-          this.attributes.health = this.attributes.maxHealth;
+          this.setAttribute('health', this.getRawAttribute('maxHealth'));
```
shawncplus: Same thing here

bundles/core-combat/player-events.js
```diff
@@ -124,22 +128,22 @@ module.exports = (srcPath) => {
...
-            let currentPerc = Math.floor((target.attributes.health / target.attributes.maxHealth) * 100);
+            let currentPerc = Math.floor((target.getAttribute('health') / target.getAttribute('maxHealth')) * 100);
             let buf = `<bold>${target.name}</bold>: `;
             buf += '<red>[<bold>';
             buf += new Array(Math.ceil((currentPerc / 100) * percWidth)).join('#') + '|';
             buf += new Array(percWidth - Math.ceil((currentPerc  / 100) * percWidth)).join(' ');
             buf += '</bold>]</red>';
-            buf += ` <bold>${target.attributes.health}/${target.attributes.maxHealth}</bold>`;
+            buf += ` <bold>${target.getAttribute('health')}/${target.getAttribute('maxHealth')}</bold>`;
             Broadcast.sayAt(this, buf);
```
shawncplus: I use these progress bars in a few places, maybe tonight I'll refactor them

bundles/core-combat/player-events.js
```diff
@@ -164,7 +168,7 @@ module.exports = (srcPath) => {
        */
       deathblow: state => function (target) {
         this.emit('experience', LevelUtil.mobExp(target.level));
-        this.attributes.health = this.attributes.maxHealth;
+        this.setAttribute('health', this.getRawAttribute('maxHealth'));
```
shawncplus: Same `getRawAttribute` comment as above for these two

src/Npc.js
```diff
@@ -42,7 +44,7 @@ class Npc extends Character {
   hydrate(state) {
     super.hydrate(state);
 
-    this.attributes.health = this.attributes.maxHealth;
+    this.setAttribute('health', this.getRawAttribute('maxHealth'));
```
shawncplus: `getRawAttribute` -> `getAttribute`

bundles/core-combat/player-events.js
shawncplus: > I would hope there would not be any effects that modify just maxHealth and not health as well.

I could imagine a skill that applied an effect that temporarily boosted max health and left health alone in which case when you use `getRawHealth` you'd be setting their health to something less than their effective max health.

seanohue: > I could imagine a skill that applied an effect that temporarily boosted max health and left health alone in which case when you use getRawHealth you'd be setting their health to something less than their effective max health.

In that case, which of the following would we want to happen:

a) Their actual `health` is set to more than their actual `maxHealth`, which becomes evident once the effect wears off, if they have not taken damage.

b) Their actual `health` is set to more than their actual `maxHealth` and we rely on the programmer of the effect to write a handler that will check for that situation once the effect is deactivated

Also, I am fine with going either way, however, the behavior should be documented (especially in the section about effects).

shawncplus: I'm not sure how either could happen. If you're always using `getAttribute('maxHealth')` you would never be able to set `health` to more than their _effective_ max health. But regardless yeah the effect's wearoff code would have to make sure to update health to the new max health if the effect wears off and `health > maxHealth`

seanohue: I may have mistyped, it would set their actual `health` to more than their actual `maxHealth`

Effect: cast a spell to double `maxHealth`
actual `health` && effective `health`: 100
actual `maxHealth`: 100
effective `maxHealth`: 200

They appear to be at 50% health now.

They use a healing potion to boost their health up to the new effective maximum, assuming use of `getAttribute('maxHealth')`:

actual `health`: 200
effective `health`: 200
actual `maxHealth`: 100
effective `maxHealth`: 200

They now appear to have 100% health.

The effect wears off. If the effect's deactivation does not account for `health` and merely reverts `maxHealth`, then we end up with this scenario:

actual `health`: 200
effective `health`: 200
actual `maxHealth`: 100
effective `maxHealth`: 100

They now appear to have 200% health and become demi-gods.

PR #107, opened 2017-02-08
Attribute deltas

seanohue:
Attributes are now stored and serialized as: 
```
{ 
 name: 'strength'
 base: 20
 delta: -2
} 
``` 
(for example).

The Attribute class allows an API that is abstracted away by the Character for raising and lowering attributes via the delta.

Some changes had to be made to various other parts of the code that used the old attributes API, most notably `Effects.evaluate`, `Player.interpolatePrompt`, and the combat handler. 

However, I think the code is now more robust, and allows for more flexibility and less chance of weird bugs. In addition, a lot of the math around attributes is now abstracted away by `Character.raiseAttribute`, `Character.lowerAttribute`, and `Character.setAttributeToMax`.


bundles/core-combat/player-events.js
```diff
@@ -128,22 +128,22 @@ module.exports = (srcPath) => {
           }
 
           // render health bars
-          let currentPerc = Math.floor((this.getAttribute('health') / this.getAttribute('maxHealth')) * 100);
+          let currentPerc = Math.floor((this.getAttribute('health') / this.geMaxtAttribute('health')) * 100);
```
shawncplus: typo `MaxtAttribute` -> `MaxAttribute`

data/player/Admin.json
```diff
@@ -1 +1 @@
-{"attributes":{"maxHealth":100,"health":100,"maxMana":200,"mana":200,"maxEnergy":200,"energy":200,"strength":20,"intelligence":20,"wisdom":20,"dexterity":20,"constitution":20},"experience":0,"inventory":[],"level":1,"name":"Admin","prefs":[],"questData":{"completed":[],"active":[["limbo:1",{"state":{"count":0}}]]},"room":"limbo:1","equipment":{}}
\ No newline at end of file
+{"attributes":{"health":100,"mana":200,"energy":200,"strength":20,"intelligence":20,"wisdom":20,"dexterity":20,"constitution":20},"experience":0,"inventory":[],"level":1,"name":"Admin","prefs":[],"questData":{"completed":[],"active":[["limbo:1",{"state":{"count":0}}]]},"room":"limbo:1","equipment":{}}
```
shawncplus: Is this correct? the serialized attributes here don't look like the format you described

bundles/core-combat/player-events.js
```diff
@@ -74,8 +74,8 @@ module.exports = (srcPath) => {
             const damage = Math.min(playerDamage, target.getAttribute('health'));
 
             target.emit('hit', this, damage);
-            const startingHealth = target.getRawAttribute('health');
-            target.setAttribute('health', startingHealth - damage);
+            const startingHealth = target.getAttribute('health');
```
shawncplus: can probably delete the `startingHealth` variables since they're not used anymore

bundles/core-combat/player-events.js
```diff
@@ -92,8 +92,8 @@ module.exports = (srcPath) => {
 
             this.emit('hit', target, damage);
 
-            const startingHealth = this.getRawAttribute('health');
-            this.setAttribute('health', startingHealth - damage);
+            const startingHealth = this.getAttribute('health');
```
shawncplus: ^

src/Character.js
```diff
@@ -39,28 +39,51 @@ class Character extends EventEmitter
...
+    return this.effects.evaluate(attribute);
+  }
+
+  /* Get value of attribute including changes to the attribute.
+   * @param {string} attr
+   * @return {number}
+  */
   getAttribute(attr) {
-    return this.effects.evaluateAttribute(attr);
+    return this.getMaxAttribute(attr) - this.attributes.get(attr).delta;
```
shawncplus: Shouldn't this be `+ delta`

src/Player.js
```diff
@@ -39,23 +39,22 @@ class Player extends Character {
...
-    for (const key in data) {
-      prompt = prompt.replace('%' + key, data[key]);
-    }
-
-    // replace any unknown tokens
-    prompt = prompt.replace(/%([a-z_]+?)/, '[BAD: $1]');
-
-    return prompt;
+  interpolatePrompt(promptBuilder, extraData = {}) {
+    const buildAttributeStr = attr => `${this.getAttribute(attr)}/${this.getMaxAttribute(attr)}`;
```
shawncplus: Gonna have to revisit this later, this change makes the prompt a lot less flexible because it only allows for health/energy. Will work on this tonight to bring back the token approach

data/player/Admin.json
seanohue: Ah, that's correct. However, the Attributes class will still handle Numbers in the constructor since it has to do so to handle NPC definitions and the default Player attributes as specified in ranvier.json. So, this won't blow up either way, but I could go ahead and change it to match the serialized format.

src/Player.js
seanohue: Yeah, I put a comment regarding that... I do like the idea of having the prompt built by a function, but agree that this is less flexible for the time being. I'd rather see something where you pass in a list of attributes to be reformatted as `current/max` and the promptBuilder recognizes that.

bundles/core-combat/player-events.js
seanohue: I thought I'd fixed that 😭  Going to double check.

PR #120, opened 2017-02-11
Combat rounds multi refactor

seanohue:


bundles/core-combat/player-events.js
```diff
@@ -151,6 +116,61 @@ module.exports = (srcPath) => {
...
+          attacker.combatData.lag = attacker.combatData.speed * 1000;
+        }
+
+        function handleDeath(deadEntity, killer) {
+          deadEntity.combatants.forEach(combatant => {
+            deadEntity.removeCombatant(combatant);
+            combatant.removeCombatant(deadEntity);
+          }, deadEntity);
+
+          if (Reflect.has(deadEntity, 'removePrompt')) {
```
shawncplus: I'd prefer this check to be `instanceof Player` or alternatively `TypeUtil.is(deadEntity, Promptable)` if you're only looking for the addPrompt/removePrompt

seanohue: Ah, I didn't know about the Promptable thing. Thanks.

package.json
```diff
@@ -25,6 +25,7 @@
     "optimist": "",
     "pretty-error": "^2.0.2",
     "purdy": "^2.2.0",
+    "rando-js": "^0.1.1",
```
shawncplus: So my only issue with this is that if we're going to go with implementing a random lib I'd prefer that we actually implement something that doesn't use `Math.random` whether it be Mersenne Twister or Alea or whatever I have no opinion

PR #125, opened 2017-02-16
Channel cleanup and commenting

seanohue:


src/Channel.js
```diff
@@ -5,19 +5,27 @@ const ChannelAudienceWorld = require('./ChannelAudience/World');
 const ChannelAudiencePrivate = require('./ChannelAudience/Private');
 const util = require('util');
 
+class ChannelError extends Error {}
```
shawncplus: Unless we're exporting `ChannelError` so people can catch and handle ChannelErrors specifically there's no real reason to actually use this instead of a plain `Error`

src/Channel.js
```diff
@@ -29,22 +37,20 @@ class Channel {
...
   send(state, sender, message) {
+
+    // If they don't include a message, explain how to use the channel.
     if (!message.length) {
-      Broadcast.sayAt(sender, `Channel: ${this.name}`);
-      if (this.description) {
-        Broadcast.sayAt(sender, this.description);
-      }
-      return this.showUsage(sender);
+      return this.describeSelf(sender);
```
shawncplus: :+1: good change

src/ChannelAudience/Area.js
```diff
@@ -3,17 +3,19 @@
 const ChannelAudience = require('../ChannelAudience');
 
 /**
- * Audience class representing other players in the area as the sender
+ * Audience class representing other players in the same area as the sender
  */
-class Area extends ChannelAudience {
+class AreaAudience extends ChannelAudience {
```
shawncplus: I tend to have a convention of class name = filename. If someone wants to namespace the class as `AreaAudience` they can do so during the `require()` e.g., `const AreaAudience  = require('./ChannelAudience/Area');`

src/ChannelAudience/Private.js
```diff
@@ -3,9 +3,10 @@
 const ChannelAudience = require('../ChannelAudience');
 
 /**
- * Audience class representing other players in the area as the sender
+ * Audience class representing a specific targeted player.
+ * Example: `tell` command or `whisper` command.
  */
-class Area extends ChannelAudience {
+class PrivateAudience extends ChannelAudience {
```
shawncplus: Same

src/ChannelAudience/Room.js
```diff
@@ -3,12 +3,15 @@
 const ChannelAudience = require('../ChannelAudience');
 
 /**
- * Audience class representing other players in the room as the sender
+ * Audience class representing other players in the same room as the sender
+ * Could even be used to broadcast to NPCs if you want them to pick up on dialogue,
+ * just make them broadcastables.
  */
-class Room extends ChannelAudience {
+class RoomAudience extends ChannelAudience {
```
shawncplus: Same

src/ChannelAudience/World.js
```diff
@@ -3,12 +3,12 @@
 const ChannelAudience = require('../ChannelAudience');
 
 /**
- * Audience class representing everyone in the game
+ * Audience class representing everyone in the game, except sender.
  */
-class World extends ChannelAudience {
+class WorldAudience extends ChannelAudience {
```
shawncplus: Same

src/ChannelAudience/Private.js
seanohue: I'll change it, I just noticed the copy/paste error and thought I'd make the names a bit clearer.

seanohue:
Changes made!

PR #129, opened 2017-02-17
Add blurb in README about debugging with VS Code

jackjwilliams:


jackjwilliams:
Alrighty, sounds good! Will update

PR #131, opened 2017-02-17
Feature/help files

jackjwilliams:
Added various and sundry help files. 

Added printout of aliases if a command has aliases when printing help file

### DIFF
--- a/bundles/core-commands/commands/commands.js
+++ b/bundles/core-commands/commands/commands.js
@@ -49,6 +49,12 @@ module.exports = (srcPath) => {
         }
       }
 
+      // Command Usage
+      Broadcast.sayAt(player);
+      Broadcast.sayAt(player, "<bold><white>                   Usage</bold></white>");
+      Broadcast.sayAt(player, "<bold><white>===============================================</bold></white>");
+      Broadcast.sayAt(player, "command <required> [optional]");
+      
       // end with a line break
       Broadcast.sayAt(player, '');
     }
--- a/bundles/core-commands/commands/commands.js
+++ b/bundles/core-commands/commands/commands.js
@@ -48,12 +48,6 @@ module.exports = (srcPath) => {
           Broadcast.sayAt(player, '');
         }
       }
-
-      // Command Usage
-      Broadcast.sayAt(player);
-      Broadcast.sayAt(player, "<bold><white>                   Usage</bold></white>");
-      Broadcast.sayAt(player, "<bold><white>===============================================</bold></white>");
-      Broadcast.sayAt(player, "command <required> [optional]");
       
       // end with a line break
       Broadcast.sayAt(player, '');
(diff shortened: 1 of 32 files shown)


jackjwilliams:
If I need to squash, let me know

jackjwilliams:
Note on the optional vs. required parameters, example usage: quest <log/list/complete/start> <npc> [number]

I'm using <> to indicate a required and [] to indicate optional. Could be changed to "log/list/complete/start" as no angle brackets typically indicates a required argument (ex: drop item, not drop <item>).

bundles/core-commands/commands/get.js
```diff
@@ -6,6 +6,7 @@ module.exports = (srcPath) => {
   const ItemType = require(srcPath + 'ItemType');
 
   return {
+    usage: 'get item [container]',
```
seanohue: This may need to be `get <item> [container]` if `<>` means required and `[]` means optional.

bundles/core-commands/commands/move.js
```diff
@@ -9,6 +9,7 @@
 module.exports = (srcPath) => {
   return {
     aliases: [ "go", "walk" ],
+    usage: 'move direction',
```
seanohue: `move <direction>`?

bundles/core-commands/help/quest.yml
```diff
@@ -0,0 +1,9 @@
+---
+command: quest
+body: |
+  View information about quests
+  quest log               Show current quests
+  quest list npc          List quests for a given NPC
+  quest complete number   Complete the given quest
+  quest start npc number  Start the given quest by the given NPC
+
```
seanohue: I think the format for this is fine.

src/Helpfile.js
```diff
@@ -31,11 +31,17 @@ class Helpfile {
...
     header += bar;
 
-    const formatUsageSyntax = usage => `Syntax: ${usage}\r\n\r\n`;
+    const formatHeaderItem = (item, value) => `${item}: ${value}\r\n\r\n`;
     if (this.command) {
-      header += formatUsageSyntax(state.CommandManager.get(this.command).usage);
+      let actualCommand = state.CommandManager.get(this.command);
+
+      header += formatHeaderItem('Syntax', actualCommand.usage);
+
```
shawncplus: trailing whitespace

bundles/core-commands/commands/commands.js
```diff
@@ -49,6 +49,12 @@ module.exports = (srcPath) => {
         }
       }
 
+      // Command Usage
+      Broadcast.sayAt(player);
+      Broadcast.sayAt(player, "<bold><white>                   Usage</bold></white>");
+      Broadcast.sayAt(player, "<bold><white>===============================================</bold></white>");
+      Broadcast.sayAt(player, "command <required> [optional]");
+
```
shawncplus: I think it'd be better to just put this in the `help` helpfile under where it says `help [command]` something like "Command usage will show arguments as either <required> or [optional], e.g., look [target]"

PR #132, opened 2017-02-17
Fix bug in parseDot

jackjwilliams:


src/CommandParser.js
```diff
@@ -164,15 +165,15 @@ class CommandParser {
 
       // prioritize keywords over item/player names
       if (entry.keywords && entry.keywords.indexOf(keyword) !== -1) {
-        if (encountered + 1 === findNth) {
-          encountered++;
+        encountered++;
```
jackjwilliams: Need to encounter++ once found one, not once it === findNth I think?

PR #138, opened 2017-02-18
Learning commands: added put <item> <container>

jackjwilliams:
Learning how bundles and commands work, added put <item> <container>.

Added custom attribute to item: maxItems.

Unsure if I need to create any listeners for the emitted events at the end?

### DIFF
--- a/bundles/core-commands/commands/put.js
+++ b/bundles/core-commands/commands/put.js
@@ -21,10 +21,11 @@ module.exports = (srcPath) => {
         return Broadcast.sayAt(player, "Where do you want to put it?");
       }
 
-      let fromList = player.inventory, fromArg = parts[0], item = null, toContainer = null, toArg = parts[1];
-
-      toContainer = Parser.parseDot(toArg, player.room.items);
-      item = Parser.parseDot(fromArg, fromList);
+      const fromList = player.inventory;
+      const fromArg = parts[0];
+      const toArg = parts[1];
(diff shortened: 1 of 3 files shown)


bundles/core-commands/commands/put.js
```diff
@@ -0,0 +1,55 @@
...
+      }
+
+      // put 3.foo in bar -> put 3.foo bar -> put 3.foo into bar
+      const parts = args.split(' ').filter(arg => !arg.match(/in/) && !arg.match(/into/));
+
+      if (parts.length === 1) {
+        return Broadcast.sayAt(player, "Where do you want to put it?");
+      }
+
+      let fromList = player.inventory, fromArg = parts[0], item = null, toContainer = null, toArg = parts[1];
```
seanohue: I'd suggest putting each of these on their own line if they are being initialized, possibly even their own let statement.

bundles/core-commands/commands/put.js
```diff
@@ -0,0 +1,55 @@
...
+
+      toContainer = Parser.parseDot(toArg, player.room.items);
```
seanohue: Why not just initialize `toContainer` and `item` using these functions (which will return null if the container/item are not found) and have them be constants?

jackjwilliams: Yep, and honestly once we've reached this point nothing changes (or should change) so mostly everything here can be const.

PR #140, opened 2017-02-20
Users should have to confirm passwords on account creation

seanohue:
As someone with sausage fingers who tends to fumble their password typing every now and then, I would appreciate this feature, as may others.

### DIFF
--- a/bundles/core-input-events/input-events/confirm-new-password.js
+++ b/bundles/core-input-events/input-events/confirm-new-password.js
@@ -23,7 +23,7 @@ module.exports = (srcPath) => {
           return socket.emit('new-account-password', socket, args);
         }
 
-        return socket.emit('choose-character', socket, { account: args.account });
+        return socket.emit('create-player', socket, { account: args.account });
       });
     }
   };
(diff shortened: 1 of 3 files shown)


bundles/core-input-events/input-events/confirm-new-password.js
```diff
@@ -0,0 +1,30 @@
...
+
+      socket.once('data', pass => {
+        socket.toggleEcho();
+
+        if (!args.account.checkPassword(pass.toString().trim())) {
+          write("<red>Passwords do not match.</red>\r\n");
+          return socket.emit('new-account-password', socket, args);
+        }
+
+        return socket.emit('choose-character', socket, { account: args.account });
```
shawncplus: So it looks like you updated `new-account-password` to go to `confirm-new-password` instead of `create-player` but `confirm-new-password` goes to `choose-character` instead of `create-player`. I haven't run this but just reading the code it would seem you can't create new characters

seanohue: D'oh

seanohue:
Should be fixed.

PR #141, opened 2017-02-20
Chipping in to the effects docs.

seanohue:


docs/extending/effects.md
```diff
@@ -1,16 +1,18 @@
-Effects temporarily or permanently change their target such as
+Effects temporarily or permanently change attributes of their target. For example:
```
shawncplus: Effects don't have to change attributes at all, they could just make the target say something weird every 3 minutes for example.

seanohue:
That's true, I will edit it to make clear that is one of many things that
attributes can do.


Sean O'Donohue
--
*Web Application Developer*
*GM NGI Onsite Team*

PR #144, opened 2017-02-20
Add related topics

seanohue:
Also added a keyword to `help who` since I'll want to use that as a hook to create `help search [keyword]`.


shawncplus:
Merged with #146

Issue #94, opened 2017-01-25
A Web API for building/administration


seanohue:
Building a world viewer dashboard and eventually an editor, and thus a web API is necessary. Really just putting this issue out there so it's documented.

Needs:
[  ] A POST route for authentication.
[X] A GET request for each piece of world date (WIP).
[  ] POST routes to add new entities (NPCs, Rooms, etc) for builders.

Nice-to-have:
[  ] Ability to spawn copies of mobs or items.
[  ] Ability to edit existing entities, including players.
[  ] Ability to execute server commands (such as restart) or to use admin-related chat channels directly from the dashboard.

jackjwilliams:
If there is one thing I have done before, it's build web apis. If you can give some more vision / guidance I can work on expanding the gets / posts / puts (potentially?). Specific questions / thoughts.

Based on my understanding of the architecture: 
Factories are specifically for easy creation / cloning (it does maintain it's own list)
Managers are really just for managing the list of entities (it does maintain it's own list). This is what most of the system uses.
BundleManager is the entry point that populates the managers, potentially using the factories.

The Data.js class is specifically for loading / saving player / account data.

1. I'm thinking something similar to Data.js, but specifically for Bundles, Areas, etc...
2. Do we want to support updating (puts?). Or maybe only POSTS for now?

It may be easiest to begin with areas. It should support naming (POST to create manifest.yml), and adding items / npcs / rooms. Quests are a kind of special since they are added by script.

seanohue:
Right now it basically just has the GET requests for getting data. I think that there are a few features I'd want to have in a web API:

1. Creating new instances of already-defined mobs, items, etc. Relatively easy, would just need to hook into one of the factories as you mention.
2. Adding a new definition of a room, item, mob (which would then be able to be added to the game world via API #1. Adding a lot of custom scripting is out of scope but using the already-packaged behaviors should be easy.
3. There is already a WebInterface.js for handling the GET requests and the endpoints. I'm thinking maybe it could also hook into a WebEditor or WebBuilder class that would translate/delegate the POST or PUT requests to the factories/managers, and also handle saving new definitions to the filesystem as .yml -- which begs the question, where would it go? Would the POST request also specify a bundle name?

jackjwilliams:
1. I like this idea, might be a good starting point.
2. Yep, but this also begs the questions: When do these newly created files get loaded? Upon reboot? Can it / should it hook into #113 (all these questions apply too)? 
[...]

PR #148, opened 2017-02-24
Basic API

jackjwilliams:
Notes:

@seanohue I missed your comments about the keys validation, I will have to look at that.

I'm sure we can refactor this somehow to use the Data (and Area) class (I could have created a Bundle class as well, did not notice we weren't using one as of yet). I saw that Area.js hooks into updateTick, wasn't sure how to get around that as this is strictly for file / folder changes to "build out" areas / bundles (doesn't update GameState).

js-yaml does some weird stuff when saving yaml (it formats it different than the current items / rooms / npcs yaml files). I think it's still valid yaml, but ['item1', 'item2'] gets changed to 

- item 1
- item 2

I tried changing BundleManager to allow passing in a root src and bundles folder (to allow setting up "test" environments / bundles for writing tests), but the BundleManager and all subsequent core-* bundles are dependent on the src/ path. Wasn't sure how to proceed. Will setting up "test" environments / bundles require all core-* bundles be copied?

I'm holding off on the reboot endpoints and authentication for now as this PR is growing large, and that would be a significant refactor. We still only allow localhost, 127.0.0.1 and ::1.

#94


ranvier.json
```diff
@@ -15,29 +19,35 @@
     "core-quests",
     "debug-commands"
   ],
-
+  "excludedBuilderBundles": [
```
shawncplus: This isn't necessary. `bundles` already acts as a whitelist for data loaded into the game and any data loaded into the game should be loaded into the API

src/Data.js
```diff
@@ -28,12 +28,41 @@ module.exports = {
     return parsers[ext](contents);
   },
 
+  saveFile: function(filepath, data, callback) {
+    if (!fs.existsSync(filepath)) {
+      throw new Error(`File [${filepath}] does not exist!`);
+    }
+
+    const deserializers = {
```
shawncplus: Shouldn't these be called serializers?

shawncplus: Also, you have trailing whitespace :P

shawncplus:
This is a pretty big change so I'm just kind of trusting that the nuts & bolts of this works. I have to say I'm not a fan of the whole `FileManagement`/`*Folder` naming convention as it shouldn't matter to anyone except the classes loading the bundle data themselves that data is coming from the filesystem. If you want to change that now I would appreciate it but I'm not going to raise hell over it.

src/Data.js
jackjwilliams: image

PR #152, opened 2017-02-26
Added tests and modifications to data managers

jackjwilliams:


src/DataManagement/BundleData.js
```diff
@@ -89,6 +83,23 @@ class BundleData {
     fs.writeFileSync(path.join(this.baseAreasPath, name, 'items.yml'), '');
     fs.writeFileSync(path.join(this.baseAreasPath, name, 'rooms.yml'), '');
     fs.writeFileSync(path.join(this.baseAreasPath, name, 'npcs.yml'), '');
```
seanohue: Since these 4 lines are almost the same, it could be encapsulated in a descriptive method call.

For example, 
```
writeNewAreaFile(areaName, filename, data = '') {
    fs.writeFileSync(path.join(this.baseAreasPath, areaName, filename), data);
}
```

seanohue: Then, since it looks like the goal of this is just to set up a "skeleton" area with a manifest, you could have a method called something like `createBlankArea` that takes the name, title, and suggested level and does the work of lines 77-85 in one go.

test/src/DataManagement/BundleDataManagementTests.js
```diff
@@ -0,0 +1,330 @@
...
+    fs.mkdirSync(path.join(__dirname, 'test-bundle-data'));
+    fs.mkdirSync(path.join(__dirname, 'test-bundle-data', 'bundles'));
+    dataManager.loadBundles(path.join(__dirname, 'test-bundle-data'));
+    dataManager
+      .createBundle('rooms-bundle')
+      .createArea({
+        name: 'new-area',
+        title: 'New Area',
+        suggestedLevel: '1-10'
+      });
```
seanohue: This is a pretty nice looking API in practice, good job.

test/src/DataManagement/BundleDataManagementTests.js
```diff
@@ -0,0 +1,330 @@
...
+
+      it("returns false when bundle doesn't exist", function() {
+        expect(dataManager.bundleExists('not-a-real-bundle')).to.be.false;
+      });
+    });
+
+    describe("getBundleDirectory", function() {
+      it("getBundleDirectory returns a bundle if it exists", function() {
+        expect(dataManager.getBundleDirectory('rooms-bundle')).to.not.be.null;
```
seanohue: A better test could be `expect(...).to.be.ok` -- more expressive, and makes sure it is neither null nor undefined.

test/src/DataManagement/BundleDataManagementTests.js
```diff
@@ -0,0 +1,330 @@
...
+      });
+
+      it("getBundleDirectory returns null if it doesn't exist", function() {
+        expect(dataManager.getBundleDirectory('not-a-real-bundle')).to.be.undefined;
+      });
+    });
+
+    it("createBundle creates a bundle", function() {
+      dataManager.createBundle('awesome-bundle');
+      expect(dataManager.getBundleDirectory('awesome-bundle')).to.not.be.null;
```
seanohue: Same as above with `.to.be.ok`

test/src/DataManagement/BundleDataManagementTests.js
```diff
@@ -0,0 +1,330 @@
...
+      area = bundle.createArea({
+        name: 'test-area',
+        title: 'Test Area',
+        suggestedLevel: '1-10'
+      });
+    });
+
+    it("getArea returns an area", function() {
+      const area = bundle.getArea('test-area');
+      expect(area).to.not.be.null;
```
seanohue: see above

src/DataManagement/BundleData.js
jackjwilliams: I like it! Will get an update up soon

test/src/DataManagement/BundleDataManagementTests.js
jackjwilliams: I was wondering if there was a better way, thanks!

src/DataManagement/BundleDataManager.js
```diff
@@ -33,11 +34,7 @@ class BundleDataManager {
       this.loadBundle(bundle);
     }
 
-    const watcher = chokidar.watch(this.basePath, {ignoreInitial: true});
-
-    watcher.on('addDir', newDir => {
-      this.loadBundle(path.basename(newDir));
-    });
+    // this.watch();
```
shawncplus: Is this commented out on purpose or accidental?

jackjwilliams: This was meant to be removed. I removed all file watching for now.

PR #155, opened 2017-02-27
Pvp enable

seanohue:
May add a flag to `ranvier.json` to enable/disable PvP, either in this PR or a forthcoming one, based on comments.


seanohue:
Also, added `Broadcast.atExcept` functionality, which will allow for broadcasting to everyone in a room/area except a list of broadcastables. This is good for having, as seen in this PR, messages that differ for the player doing the action, their target, and any 3rd party observers.

bundles/ranvier-combat/player-events.js
```diff
@@ -211,38 +220,54 @@ module.exports = (srcPath) => {
...
     }
 
-    deadEntity.emit('killed', target || deadEntity);
+    Broadcast.sayAtExcept(
+      (killer ? [killer, deadEntity] : deadEntity),
+      deadEntity.room,
+      othersDeathMessage);
+
+    if (killer && killer instanceof Player) {
+      Broadcast.sayAt(killer, `<bold><red>You killed ${deadEntity.name}!`);
```
shawncplus: This should be in the `deathblow` event

src/Broadcast.js
```diff
@@ -28,6 +28,26 @@ class Broadcast {
     });
   }
 
+  static atExcept(excludes, source, message, wrapWidth, useColor, formatter) {
```
shawncplus: I'd prefer if this keep the source, message arguments first so its prototype matches more closely the other Broadcast methods. So: `source, message, excludes`

src/Broadcast.js
```diff
@@ -38,6 +58,12 @@ class Broadcast {
     });
   }
 
+  static sayAtExcept(excludes, source, message, wrapWidth, useColor, formatter) {
```
shawncplus: Same as above

bundles/ranvier-combat/commands/kill.js
```diff
@@ -16,26 +17,35 @@ module.exports = (srcPath) => {
...
+      ];
+
+      const target = Parser.parseDot(args, possibleTargets);
+      if (!target) {
         return Broadcast.sayAt(player, "They aren't here.");
       }
 
-      if (npc.hasBehavior('pacifist')) {
-        return Broadcast.sayAt(player, `${npc.name} is a pacifist and will not fight you.`);
+      if (target.hasBehavior && target.hasBehavior('pacifist')) {
```
shawncplus: I'd prefer this checked `target instanceof Npc` instead of checking for existence of hasBehavior. If you want to be _really_ cool you could make this easier for everyone and add a `isNpc()` method to the base `Character` class which returns false by default but `Npc` overrides to return true.

src/Broadcast.js
seanohue: My rationale was that it reads more fluidly to be like, at, _Except_ (the thing modified by the word "except"), rest of arguments. That said, it does make sense to keep a similar signature, so I'll go ahead and modify it.

bundles/ranvier-combat/player-events.js
seanohue: 👍

bundles/ranvier-combat/commands/kill.js
seanohue: Gotcha. I've already changed it locally to use `instanceof`. I'll take the isNpc thing into consideration, though. Might make the API nicer so we don't have to require Player or Npc everywhere just to do type checks.

src/Broadcast.js
seanohue: Done

seanohue:
Decided to add the `pvp` command as well as the `isNpc` flag on all characters.

src/Broadcast.js
```diff
@@ -28,6 +28,26 @@ class Broadcast {
     });
   }
 
+  static atExcept(source, message, excludes = [], wrapWidth, useColor, formatter) {
```
shawncplus: excludes shouldn't be optional, either they're excluding a list or they're not. And just as a general rule it's bad practice to have optional parameters between required parameters

seanohue: Should be fixed now, you're right that it is pretty goofy to have it be optional.

PR #158, opened 2017-03-02
Decay items

seanohue:
Right now this has the most basic functionality. 

The config takes a duration, in seconds, that the item will take to decay.
Once the item is spawned, `decayEnd` is set. Halfway to that time, the item's description will change to indicate that it is rotting. When `decayEnd` is reached, the item (and any of its contents, if it is a container) are removed.

As an addition, items now receive the `updateTick` event automatically, and Inventories will put a reference back to themselves onto any items they contain.

### DIFF
--- a/bundles/ranvier-areas/areas/limbo/items.yml
+++ b/bundles/ranvier-areas/areas/limbo/items.yml
@@ -10,6 +10,8 @@
   roomDesc: "A moldy slice of cheese"
   keywords: [ "slice", "cheese", "moldy" ]
   description: "A yellow, slightly moldy slice of cheese. Only a rat could find this appetizing."
+  behaviors:
+    decay: 300
 - id: 3
   type: CONTAINER
   name: "Wooden Chest"
--- a/bundles/ranvier-areas/areas/limbo/items.yml
+++ b/bundles/ranvier-areas/areas/limbo/items.yml
@@ -11,7 +11,8 @@
   keywords: [ "slice", "cheese", "moldy" ]
   description: "A yellow, slightly moldy slice of cheese. Only a rat could find this appetizing."
   behaviors:
-    decay: 300
+    decay:
+      duration: 300
 - id: 3
   type: CONTAINER
   name: "Wooden Chest"
--- a/bundles/ranvier-areas/areas/limbo/items.yml
+++ b/bundles/ranvier-areas/areas/limbo/items.yml
@@ -12,7 +12,7 @@
   description: "A yellow, slightly moldy slice of cheese. Only a rat could find this appetizing."
   behaviors:
     decay:
-      duration: 300
+      duration: 30
 - id: 3
   type: CONTAINER
   name: "Wooden Chest"
--- a/bundles/ranvier-areas/areas/limbo/items.yml
+++ b/bundles/ranvier-areas/areas/limbo/items.yml
@@ -12,7 +12,7 @@
   description: "A yellow, slightly moldy slice of cheese. Only a rat could find this appetizing."
   behaviors:
     decay:
-      duration: 30
+      duration: 240
 - id: 3
   type: CONTAINER
   name: "Wooden Chest"
--- a/bundles/ranvier-combat/behaviors/item/decay.js
+++ b/bundles/ranvier-combat/behaviors/item/decay.js
@@ -39,7 +39,7 @@ module.exports = srcPath => {
         }
 
         if (type === ItemType.CONTAINER && this.inventory) {
-          this.inventory.forEach(item => destroyItem(item));
+          this.inventory.forEach(item => destroyItem(state, item));
         }
 
         if (belongsTo) {
--- a/bundles/ranvier-combat/behaviors/item/decay.js
+++ b/bundles/ranvier-combat/behaviors/item/decay.js
@@ -15,33 +15,34 @@ module.exports = srcPath => {
 
         if (decayEnd) {
           if (decayEnd < now) {
-            destroyItem(state, this);
+            this.emit('decay');
           } else {
+            //TODO: Set some kind of metadata.
             checkForHalfRotted(this, duration, now);
           }
         } else {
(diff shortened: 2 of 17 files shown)


bundles/ranvier-areas/areas/limbo/items.yml
```diff
@@ -10,6 +10,9 @@
   roomDesc: "A moldy slice of cheese"
   keywords: [ "slice", "cheese", "moldy" ]
   description: "A yellow, slightly moldy slice of cheese. Only a rat could find this appetizing."
+  behaviors:
```
shawncplus: The way this currently works would break the fetch quest. Fetch quest listens for the 'drop' event so if decay bypasses `drop` by directly removing the item the drop event won't get fired so the quest won't be updated with the fact that the player no longer has the item

bundles/ranvier-combat/behaviors/item/decay.js
```diff
@@ -0,0 +1,58 @@
...
+          decayStarted = true;
+          decayEnd = Date.now() + duration;
+        }
+      }
+    }
+  };
+
+  function destroyItem(state, rottedItem) {
+    Logger.verbose(`${rottedItem.id} has decayed.`);
+    state.ItemManager.remove(rottedItem);
```
shawncplus: Is this removing it from the player's inventory as well?

src/Inventory.js
```diff
@@ -7,6 +7,7 @@ class Inventory extends Map {
   addItem(item) {
     this.set(item.uuid, item);
     item.isHeld = true;
+    item.container = this;
```
shawncplus: I don't... like this. `container` is supposed to be an indicator of the _item_ it's currently contained in, not a reference to an `Inventory` object.

bundles/ranvier-combat/behaviors/item/decay.js
```diff
@@ -0,0 +1,58 @@
...
+    if (rottedItem.type === ItemType.CONTAINER) {
+      rottedItem.inventory.forEach(item => state.ItemManager.removeItem(item));
+    }
+
+    if (rottedItem.container) {
+      rottedItem.container.removeItem(this);
+    }
+  }
+
+  function checkForHalfRotted(item, duration, now) {
```
shawncplus: Personally I'd prefer that the `look` command just be updated to check for the `decay` behavior and show the _exact_ time remaining. I have a strong distaste for MUDs that hide numbers when there is no good reason

bundles/ranvier-areas/areas/limbo/items.yml
seanohue: Weird (though, it makes sense). 

So, maybe there should be a decay event as well, or should force the item to be dropped by emitting the drop event?

src/Inventory.js
seanohue: Hmm. This was mainly a solution to the problem of having the `Inventory` hold a reference to an item that may have already decayed. I am not sure how else this would be solved, though it could just have a reference back to the containing item then decay could call `this.container.inventory.removeItem`

bundles/ranvier-combat/behaviors/item/decay.js
seanohue: Not until later on, but there's an issue with this due to the behavior noted below, where the item now has a reference to the inventory it is a part of.

src/Inventory.js
seanohue: Any suggestions on how to resolve this? I thought the inventory reference would be okay since it would work for both containers and for items in a player's inventory but if nothing else it should have a different name.

src/Inventory.js
```diff
@@ -6,17 +6,15 @@ class Inventory extends Map {
    */
   addItem(item) {
     this.set(item.uuid, item);
-    item.isHeld = true;
-    item.container = this;
+    item.belongsTo = this;
```
shawncplus: This is still wrong, heh. `belongsTo` should be equal to the _entity_ (either Character or Item) it belongs to. Not the actual Inventory instance. Otherwise you just kick the can down the road because sure you have a pointer to the inventory it belongs to but you can't actually get a reference to entity that the inventory belongs to. So you could never, for example, get the name of the player that is carrying an item simply by having the item.

seanohue: Hrm. That begs the question, how do I get that from inside this Inventory object? Should the Inventory have a reference to the thing it belongs to as well?

shawncplus: You don't. Both Item and Character have their own addItem/removeItem. You shouldn't ever need to directly access the `inventory` of an entity.

seanohue: me right now

bundles/ranvier-combat/behaviors/item/decay.js
seanohue: I think I finally have this part correct?

bundles/ranvier-combat/behaviors/item/decay.js
```diff
@@ -0,0 +1,98 @@
...
+      if (owner instanceof Player) {
+        Broadcast.sayAt(owner, `Your ${item.name} has rotted away!`);
+        break;
+      } else {
+        owner = owner.belongsTo;
+      }
+    }
+  }
+
+  function getTimeUntilDecay(now) {
```
shawncplus: This can be replaced by `humanize-duration`. See `ranvier-effects/commands/effects.js`

shawncplus: If anything perhaps this isn't even needed, and `humanize-duration` can just be used inside `look`

seanohue: You mean I handrolled my own time humanizer for no reason? :1st_place_medal: 

Kidding, I'll check it out and see if this can be swapped for it.

seanohue: Updooted 🎷

Issue #121, opened 2017-02-12
Player preferences


shawncplus:
There needs to be some kind of `options` or `config` command to let players have options like their default combat prompt, autoloot, autosac, brief, etc., essentially an arbitrary map of options.

PR #159, opened 2017-03-02
New command: config for #121

jackjwilliams:
Player preferences for #121 

Current options: brief and autoloot. autoloot to be used with loot tables once it's complete (or is it already complete?).

Brief simply displays the room title instead of the whole description.

### DIFF
--- a/bundles/ranvier-commands/commands/config.js
+++ b/bundles/ranvier-commands/commands/config.js
@@ -1,0 +1,89 @@
+'use strict';
+
+module.exports = (srcPath) => {
+  const Broadcast = require(srcPath + 'Broadcast');
+  const Parser = require(srcPath + 'CommandParser').CommandParser;
+  const Data = require(srcPath + 'Data');
+
+  return {
+    usage: 'config <set/list> [setting] [value]',
+    command: (state) => (args, player) => {
+      args = args.trim();
+
+      if (!args.length) {
+        Broadcast.sayAt(player, 'Configure what?');
+        return state.CommandManager.get('help').execute('config', player);
+      }
+
+      const possibleCommands = ['set', 'list'];
+      const splitArgs = args.split(' ');
+      const command = splitArgs[0];
+
+      if (possibleCommands.indexOf(command) === -1) {
+        Broadcast.sayAt(player, `<red>Invalid config command: ${command}</red>`);
+        return state.CommandManager.get('help').execute('config', player);
+      }
+
+      if (command === 'list') {
+        return listCurrentConfiguration();
+      }
+
+      if (splitArgs.length === 1) {
+        Broadcast.sayAt(player, 'Set what?');
+        return state.CommandManager.get('help').execute('config', player);
+      }
+
+      const configToSet = splitArgs[1];
+      const possibleSettings = ['brief', 'autoloot'];
+
+      if (possibleSettings.indexOf(configToSet) === -1) {
+        Broadcast.sayAt(player, `<red>Invalid setting: ${configToSet}. Possible settings: ${possibleSettings.join(', ')}`);
+        return state.CommandManager.get('help').execute('config', player);
+      }
+
+      const valueToSet = splitArgs[2];
+
+      if (!valueToSet) {
+        Broadcast.sayAt(player, `<red>What value do you want to set for ${configToSet}?</red>`);
+        return state.CommandManager.get('help').execute('config', player);
+      }
+
+      const possibleValues = [];
+
+      const trueFalse = ['true', 'false'];
+      possibleValues['brief'] = trueFalse;
+      possibleValues['autoloot'] = trueFalse;
+
+      if (possibleValues[configToSet].indexOf(valueToSet) === -1) {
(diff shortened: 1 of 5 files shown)


bundles/ranvier-commands/commands/config.js
```diff
@@ -0,0 +1,89 @@
...
+      const valueToSet = splitArgs[2];
+
+      if (!valueToSet) {
+        Broadcast.sayAt(player, `<red>What value do you want to set for ${configToSet}?</red>`);
+        return state.CommandManager.get('help').execute('config', player);
+      }
+
+      const possibleValues = [];
+
+      const trueFalse = ['true', 'false'];
```
shawncplus: This is pedantic, but on/off is much more natural and less coder-specific language

bundles/ranvier-commands/commands/look.js
```diff
@@ -12,7 +12,7 @@ module.exports = (srcPath) => {
 
     // Render the room
     Broadcast.sayAt(player, room.title);
-    Broadcast.sayAt(player, room.description, 80);
+    Broadcast.sayAt(player, player.config.brief === true ? `[Brief] ${room.title}` : room.description, 80);
```
shawncplus: Not sure this is necessary

src/Player.js
```diff
@@ -43,6 +43,7 @@ class Player extends Character {
     // WARNING: values must be JSON.stringify-able
     this.metadata = data.metadata || {};
     this.playerClass = null;
+    this.config = data.config || {};
```
shawncplus: Unnecessary, this is basically what `meta` is for. You can just use a `config` key inside of meta, e.g., "config.autoloot" (meta key supports dot notation)

bundles/ranvier-commands/commands/config.js
```diff
@@ -0,0 +1,89 @@
...
+      if (!args.length) {
+        Broadcast.sayAt(player, 'Configure what?');
+        return state.CommandManager.get('help').execute('config', player);
+      }
+
+      const possibleCommands = ['set', 'list'];
+      const splitArgs = args.split(' ');
+      const command = splitArgs[0];
+
+      if (possibleCommands.indexOf(command) === -1) {
```
seanohue: With ES6, this can be shortened to `if (possibleCommands.includes(command)) { ... `

bundles/ranvier-commands/commands/config.js
```diff
@@ -0,0 +1,89 @@
...
+    command: (state) => (args, player) => {
+      args = args.trim();
+
+      if (!args.length) {
+        Broadcast.sayAt(player, 'Configure what?');
+        return state.CommandManager.get('help').execute('config', player);
+      }
+
+      const possibleCommands = ['set', 'list'];
+      const splitArgs = args.split(' ');
```
seanohue: Since you end up using up to three args below, and naming them, you could do something like this here:
```
const [ command, configToSet, valueToSet ] = args.split(' ');
```

Then, instead of checking for `splitArgs.length` you could check for `configToSet`.

bundles/ranvier-commands/commands/config.js
seanohue: Like I said in the slack, I'd suggest something like the following:
```
const onOff = {
 on: true,
 off: false
};```

That way, you don't have to use `===` later on to get a boolean, you just match against the keys in this onOff object and get the appropriate boolean value.

File: bundles/ranvier-commands/commands/config.js (hunk 5 above)
Comment (jackjwilliams) [inline bundles/ranvier-commands/commands/config.js]: That would be much cleaner, thanks!

File: bundles/ranvier-commands/commands/config.js [hunk 6]
```diff
@@ -0,0 +1,74 @@
...
+
+module.exports = (srcPath) => {
+  const Broadcast = require(srcPath + 'Broadcast');
+  const Parser = require(srcPath + 'CommandParser').CommandParser;
+  const Data = require(srcPath + 'Data');
+
+  return {
+    usage: 'config <set/list> [setting] [value]',
+    command: (state) => (args, player) => {
+      args = args.trim();
```
shawncplus: This line isn't necessary, some old commands have it because they were written before a change to make this unnecessary :P

bundles/ranvier-commands/commands/config.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+module.exports = (srcPath) => {
+  const Broadcast = require(srcPath + 'Broadcast');
+  const Parser = require(srcPath + 'CommandParser').CommandParser;
+  const Data = require(srcPath + 'Data');
+
+  return {
+    usage: 'config <set/list> [setting] [value]',
```
shawncplus: Maybe while you're in here add aliases `toggle`, `options`, `set`

src/Player.js
```diff
@@ -176,6 +176,10 @@ class Player extends Character {
     Data.save('player', this.name, this.serialize(), callback);
   }
 
+  setConfig(config) {
```
shawncplus: This isn't used anymore

PR #162, opened 2017-03-03
Playing with some formatting and printing a directions map

jackjwilliams:
This is why I should stop coding after 11pm. Playing with some formatting and added a directional map.

directions

### DIFF
--- a/bundles/ranvier-commands/commands/look.js
+++ b/bundles/ranvier-commands/commands/look.js
@@ -55,6 +55,8 @@ module.exports = (srcPath) => {
     const [ line1, line2, line3 ] = getCompass(player);
 
     // Render the room
+    //The top line has 99 characters of ANSI coloring (114(COLORS) - 2(NW) - 1(N) - 2(NE) - 5(SPACING) - 5(SPACING))
+    //The 164 accounts for this
     Broadcast.sayAt(player, `<yellow><bold>${room.title}</bold></yellow>` + leftPad(line1, 164 - room.title.length));
     Broadcast.sayAt(player, '--------------------------------------------' + leftPad(line2, 90));
     Broadcast.sayAt(player, leftPad(line3, 166));
--- a/bundles/ranvier-commands/commands/look.js
+++ b/bundles/ranvier-commands/commands/look.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const leftPad = require('left-pad');
+
 module.exports = (srcPath) => {
   const Broadcast = require(srcPath + 'Broadcast');
   const CommandParser = require(srcPath + 'CommandParser').CommandParser;
@@ -7,7 +9,7 @@ module.exports = (srcPath) => {
   const ItemType = require(srcPath + 'ItemType');
   const Logger = require(srcPath + 'Logger');
 
-  function getDirectionsMap(player) {
+  function getCompass(player) {
     const room = player.room;
 
     const exitMap = new Map();
@@ -50,12 +52,12 @@ module.exports = (srcPath) => {
   function lookRoom(state, player) {
     const room = player.room;
 
-    const [ line1, line2, line3 ] = getDirectionsMap(player);
+    const [ line1, line2, line3 ] = getCompass(player);
 
     // Render the room
-    Broadcast.sayAt(player, Broadcast.pad(new Array(90).join(' '), `<yellow><bold>${room.title}</bold></yellow>`) + line1);
-    Broadcast.sayAt(player, Broadcast.pad(new Array(60).join(' '), '--------------------------------------------') + line2);
-    Broadcast.sayAt(player, Broadcast.pad(new Array(60).join(' '), ' ') + line3);
+    Broadcast.sayAt(player, `<yellow><bold>${room.title}</bold></yellow>` + leftPad(line1, 164 - room.title.length));
+    Broadcast.sayAt(player, '--------------------------------------------' + leftPad(line2, 90));
+    Broadcast.sayAt(player, leftPad(line3, 166));
     Broadcast.sayAt(player, room.description, 80);
     Broadcast.sayAt(player, '');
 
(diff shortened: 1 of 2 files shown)


src/Broadcast.js
```diff
@@ -137,6 +137,16 @@ class Broadcast {
     out = out.join('\r\n').replace('<NEWLINE>', '\r\n');
     return out;
   }
+
+  static pad(pad, str, padLeft) {
```
shawncplus: This is a slightly strange API for a pad function. What's more, we already have a `left-pad` package imported that you can use. https://www.npmjs.com/package/left-pad

bundles/ranvier-commands/commands/look.js
```diff
@@ -7,11 +7,55 @@ module.exports = (srcPath) => {
...
+    const line2 = `<yellow><bold>${W}</bold></yellow> <---(M)---> <yellow><bold>${E}</bold></yellow>`;
+    const line3 = `<yellow><bold>${SW}</bold></yellow>     <yellow><bold>${S}</bold></yellow>     <yellow><bold>${SE}</bold></yellow>\r\n`;
+
+    return [line1, line2, line3];
+  }
+
   function lookRoom(state, player) {
     const room = player.room;
 
+    const [ line1, line2, line3 ] = getDirectionsMap(player);
```
shawncplus: This is a good opportunity to add a player pref for this too, this feature is generally called a `compass` so off by default but `config compass on` to enable it

src/Broadcast.js
seanohue: I, too, have been guilty of trying to build a better left-pad.

bundles/ranvier-commands/commands/look.js
jackjwilliams: Yep, good idea. Once we get these two branches merged together I'll add the config.

bundles/ranvier-commands/commands/look.js
```diff
@@ -50,12 +52,12 @@ module.exports = (srcPath) => {
...
     const room = player.room;
 
-    const [ line1, line2, line3 ] = getDirectionsMap(player);
+    const [ line1, line2, line3 ] = getCompass(player);
 
     // Render the room
-    Broadcast.sayAt(player, Broadcast.pad(new Array(90).join(' '), `<yellow><bold>${room.title}</bold></yellow>`) + line1);
-    Broadcast.sayAt(player, Broadcast.pad(new Array(60).join(' '), '--------------------------------------------') + line2);
-    Broadcast.sayAt(player, Broadcast.pad(new Array(60).join(' '), ' ') + line3);
+    Broadcast.sayAt(player, `<yellow><bold>${room.title}</bold></yellow>` + leftPad(line1, 164 - room.title.length));
```
shawncplus: this seems very magical, what is 164?

bundles/ranvier-commands/commands/look.js
```diff
@@ -1,17 +1,63 @@
...
   const CommandParser = require(srcPath + 'CommandParser').CommandParser;
   const Item = require(srcPath + 'Item');
   const ItemType = require(srcPath + 'ItemType');
   const Logger = require(srcPath + 'Logger');
 
+  function getCompass(player) {
+    const room = player.room;
+
+    const exitMap = new Map();
+    exitMap.set('east', 'E');
```
seanohue: It's possible to chain `Map.set` to make this a bit less repetitive. 

In addition, while a `Map` is a good candidate for containing values that you won't know at runtime, we know all of the cardinal directions at runtime. So this might not be the best choice here? Not sure though since custom exits could exist as well -- but then how would they be rendered by the compass? Or would they?

seanohue: As an example of what I mean by chaining:
```
exitMap.set('east', 'E')
    .set('west'. 'W')
    .set('south', 'S')
// ... and so on
```

bundles/ranvier-commands/commands/look.js
```diff
@@ -1,17 +1,63 @@
...
+    const line2 = `<yellow><bold>${W}</bold></yellow> <---(M)---> <yellow><bold>${E}</bold></yellow>`;
+    const line3 = `<yellow><bold>${SW}</bold></yellow>     <yellow><bold>${S}</bold></yellow>     <yellow><bold>${SE}</bold></yellow>\r\n`;
+
+    return [line1, line2, line3];
+  }
+
   function lookRoom(state, player) {
     const room = player.room;
 
+    const [ line1, line2, line3 ] = getCompass(player);
```
seanohue: Overall, this is a pretty creative way of handling a surroundings-map rendering.

bundles/ranvier-commands/commands/look.js
jackjwilliams: That's true - an array may make more sense here. I don't think custom exits could exist with this, they would not be rendered by the compass currently. Would have to be accounted for.

bundles/ranvier-commands/commands/look.js
seanohue: If we were doing fancy-pants telnet negotiation, this might be a good place to know the maximum width supported by a client. Is that even a thing. I don't know telnet.

bundles/ranvier-commands/commands/look.js
seanohue: I'd be  fine with custom exits just not being handled in `ranvier-` bundles. 

An array would probably be a 2D array of:
```
[
  ["NW", "N" "NE"]
  ["W",     "_",  "E"]
//you get it by now I hope
```

But then might have to have a way to translate the NE and such to different values, or the other way around.

jackjwilliams: Yeah I see what you are saying, but how can that map from a rooms.yml 'east' to a compass map 'E'?

seanohue: Yeah, I overthought that. Thanks 🍻 .

Probably just fine to do 
```
{
  east: "E",
  west: "W",
// and so on
}
```
in that case.
```

PR #174, opened 2017-03-05
Add find owner to item class

seanohue:
Not sure why my git history is jacked and those two extra commits keep showing up, but they do.

### DIFF
--- a/bundles/ranvier-combat/behaviors/item/decay.js
+++ b/bundles/ranvier-combat/behaviors/item/decay.js
@@ -40,7 +40,6 @@ module.exports = srcPath => {
           const owner = this.findOwner();
           if (owner) {
             Broadcast.sayAt(owner, `Your ${this.name} has rotted away!`);
-            Broadcast.prompt(owner);
           }
           belongsTo.removeItem(this);
         }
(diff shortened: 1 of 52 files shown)


bundles/ranvier-combat/behaviors/item/decay.js
```diff
@@ -37,8 +37,12 @@ module.exports = srcPath => {
         }
 
         if (belongsTo) {
+          const owner = this.findOwner();
+          if (owner) {
+            Broadcast.sayAt(owner, `Your ${this.name} has rotted away!`);
+            Broadcast.prompt(owner);
```
shawncplus: I generally only prompt the player if their stats have changed and need to see them or if the output actually requires feedback.

seanohue: Ah, gotcha. It looked weird to me during testing to have it say that and then not prompt me. But I can remove it.

shawncplus: Yeah I've generally been following the model I mentioned it makes the game a lot less spammy. Especially if you're just doing stuff like chatting

seanohue: Updated ☝️

Issue #166, opened 2017-03-05
Equipment stats


shawncplus:
Equipment and weapons should increase/decrease players stats. Weapons additionally should have speed and damage.

* Update `equip`/`unequip` to raise/lower stats with `Attribute.setBase()` 
* Update combat bundle player-events to look for a player's equipped weapon/strength for damage instead of just a static `rand(0, 20)`
* I think it should apply to both NPCs and players. I could imagine mobs that wield powerful weapons and need to be disarmed to be dealt with safely.

PR #181, opened 2017-03-05
for #166 equipment now has stats, weapons have damage

shawncplus:
As mentioned in slack this update essentially makes the core opinionated about stats/damage calculations. I think that's okay. That's one of the things that's so core to how the game runs that it's hard to make it super abstract and have people just put something in a config file without it being incredibly unwieldy.

* Weapons now have damage/speed by way of `minDamage/maxDamage/speed` on `properties
* All items can now modify the player's attributes with `properties.stats` on the definition (see `items.yml`
* Skills can make use of the weapon damage calculate with `character.calculateWeaponDamage()`
* A bunch of attribute formulas were ripped from WoW simulation data


bundles/ranvier-commands/commands/wear.js
```diff
@@ -13,33 +14,63 @@ module.exports = (srcPath) => {
...
+        return say(player, "You can't use that yet.");
       }
 
       try {
         player.equip(item);
       } catch (err) {
         if (err instanceof EquipSlotTakenError) {
           const conflict = player.equipment.get(item.slot);
-          return Broadcast.sayAt(player, `You will have to remove ${conflict.display} first.`);
+          return say(player, `You will have to remove ${conflict.display} first.`);
```
jackjwilliams: Might be nice if there was a switch to auto-replace any worn equipment when equipping new eq.

data/player/Admin.json
```diff
@@ -1,62 +1,34 @@
...
-    ],
-    [
-      "constitution",
-      {
-        "delta": 0,
-        "base": 20
-      }
-    ]
-  ],
+  "attributes": {
```
jackjwilliams: I'm sure you've noticed this by now, I've changed some of the JSON.stringify(data) to JSON.stringify(data, null, 2) for better json formatting (so we can read it easier)

ranvier
```diff
@@ -19,9 +19,9 @@ const net = require('net');
 const commander = require('commander');
 const wrap = require('wrap-ansi');
 const argv = require('optimist').argv;
-const longjohn = require('longjohn');
-const pe = require('pretty-error').start();
-pe.skipNodeFiles(); // Ignore native node files in stacktrace.
+//const longjohn = require('longjohn');
+//onst pe = require('pretty-error').start();
```
jackjwilliams: Commented lines need removal?

src/Character.js
```diff
@@ -60,6 +61,26 @@ class Character extends EventEmitter
...
+    // health and attackpower are calculated based off other stats
+    // NOTE: This is an _example_ implementation based off WoW formulas
+    switch (attribute.name) {
+      case 'health':
+        attribute.setBase(
+          this.getMaxAttribute('stamina') * healthPerStamina(this.level)
+        );
+        break;
+      case 'attackpower':
+        attribute.setBase(this.getMaxAttribute('strength'));
```
jackjwilliams: Interesting, why is a getter (getMaxAttribute) setting things?

src/Character.js
```diff
@@ -325,3 +382,159 @@ class Character extends EventEmitter
...
+    99: [ 1405, 859, 859, 687 ],
+    100: [ 1455, 889, 890, 711 ],
+    101: [ 3971, 2426, 2429, 1941 ],
+    102: [ 4547, 2778, 2781, 2222 ],
+    103: [ 5201, 3178, 3182, 2542 ],
+    104: [ 5895, 3602, 3606, 2881 ],
+    105: [ 6478, 3958, 3962, 3165 ],
+    106: [ 7105, 4341, 4346, 3472 ],
+    107: [ 7740, 4729, 4734, 3782 ],
+    108: [ 8491, 5188, 5194, 4149 ],
```
jackjwilliams: Might be nice if these and Armor reduction constants were configurable (one day).

src/Effect.js
```diff
@@ -147,8 +147,15 @@ class Effect extends EventEmitter {
    * @return {number} attribute modified by effect
    */
   modifyAttribute(attrName, currentValue) {
-    const modifier = (this.modifiers.attributes[attrName] || (_ => _)).bind(this);
-    return modifier(currentValue);
+    let modifier = _ => _;
```
jackjwilliams: While you couldn't possibly answer all questions I had, what in the world is let modifier = _ => _ doing?

src/Character.js
seanohue: That's not exactly what's going on. Nested function calls may make it harder to read, but here is what's going on:

```
const strength = this.getMaxAttribute('strength'); // This will return the value of their strength attribute as modified by effects. E.g., 18.
attribute.setBase(strength); // This will set the current attribute (attackpower?) to 18.
```

src/Effect.js
seanohue: It's the identity function, could also be written as `x => x`. It returns whatever it gets. Handy for two reasons: attribute modification in effects works as a reducer, so if nothing else, the effect is just going to keep the attribute the same. If it were undefined that could cause issues (the infamous `undefined is not a function`.

src/Character.js
seanohue: Agreed. These were taken pretty much verbatim from WoW which is fine for the defaults, but I'd like to see it pulled out into a bundle somehow.

src/Character.js
```diff
@@ -325,3 +382,159 @@ class Character extends EventEmitter
...
+
+function baseAttributeByLevel(name, level) {
+  var index = ({
+    strength: 0,
+    agility: 1,
+    stamina: 2,
+    intellect: 3,
+  })[name];
+
+  var bases = {
```
seanohue: ಠ_ಠ `var`, my mortal nemesis

jackjwilliams: All your bases var belong to us

src/Character.js
```diff
@@ -60,6 +61,26 @@ class Character extends EventEmitter
    */
   getMaxAttribute(attr) {
     const attribute = this.attributes.get(attr);
+
+    // health and attackpower are calculated based off other stats
+    // NOTE: This is an _example_ implementation based off WoW formulas
+    switch (attribute.name) {
```
seanohue: This is a cool way to set up "calculated attributes" (those based off of other stats).

ranvier
seanohue: I personally like both pretty error and long stack traces, but they might not be necessary now, especially with Winston, and if we do decide to consider using them, they should probably be extracted to the `Logger` class.

src/Effect.js
jackjwilliams: Ahhh ok I see now. You are setting a default, otherwise if the modifiers attributes is a function you defer to it. Pretty sweet.

seanohue: Yeah, looks like it's similar to the previous setup, except that `modifiers.attributes` can be either a function, that probably contains a switch/case to determine how to modify the attributes, or it can be an object with keys mapped to functions that modify the attributes. Pretty nifty, actually. Is this all documented in the effects docs?

ranvier
shawncplus: This was accidental, these two break stack traces when using a real debugger so I disable them when deving

docs/extending/effects.md
```diff
@@ -144,6 +144,25 @@ module.exports = srcPath => {
...
+      this pattern which is used when you want a base effect that could apply
+      to multiple attributes.  See the `equip.js` effect for an example
+
+      state: {
+        stat: 'strength',
+        bonus: 5
+      },
+
+      attributes: function (attribute, current) {
+        if (attribute !== this.state.state) {
```
seanohue: Should be `this.state.stat`.

src/AttributeUtil.js
```diff
@@ -0,0 +1,169 @@
+'use strict';
```
seanohue: I like this update.

PR #183, opened 2017-03-05
Admin role flag

seanohue:


bundles/ranvier-input-events/input-events/commands.js
```diff
@@ -51,12 +55,18 @@ module.exports = (src) => {
...
             }
           }
-        } catch (e) {
-          if (e instanceof InvalidCommandError) {
-            Broadcast.sayAt(player, "Huh?");
-          } else {
-            Logger.error(e);
+        } catch (error) {
+          switch(true) {
+            case error instanceof InvalidCommandError:
```
shawncplus: I love this technique, a lot of people are surprised it even works lol

seanohue: Yeah, I tried using `switch(error[Symbol.species])` and matching on that... but for Error that ends up being undefined (same with other built-in classes, I think). I like switch(true) over a bunch of if/else statements, though...

PR #191, opened 2017-03-08
Grouping commands/functionality

shawncplus:
There is now a group command which has create/invite/join/decline/disband.

Group members in the same room:

* get credit for kills via the 'deathblow' event so they will get credit for quest kills, etc.
* get any currency dropped by a lootable npc distributed evenly between them
* See combat messages for other group members in combat

There's also a new channel: `gtell` to let group members communicate across the game

### DIFF
--- a/bundles/ranvier-areas/areas/limbo/quests.js
+++ b/bundles/ranvier-areas/areas/limbo/quests.js
@@ -13,13 +13,14 @@ module.exports = (srcPath) => {
     1: {
       config: {
         title: "A Journey Begins",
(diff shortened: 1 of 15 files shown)


samjaninf:
I'm probably late to the party but just curious if there is any limitation on that exp thing? As in do party members get exp is they are across the world? Or is it just limited to if they partook in the fight? Or maybe if they are in the same room or the same area or some such.

shawncplus:
It only applies to members in the same room

Issue #172, opened 2017-03-05
Basic shop functionality


seanohue:
The logic for an in-game shop could be based around the following:

- A `shop` behavior that could be attached to an NPC, a room, or an item (for example, a vending machine). This would be configured with IDs of the items available, the quantity available (defaulting to infinite), and the cost.

- A `list` command that would list all items available at the shop if applicable.

- A `buy` command that would allow purchase if the player has enough currency. `buy <item or #>`

- A `sell` command that would allow the player to sell one of their items for currency if the shop allows. A nice-to-have would be having the shop keep the item in its inventory, at least until reboot.

One prerequisite for this is currency.

shawncplus:
OTOMH I'd say this could be done with a behavior like lootable where the config is just a list of items for sale and their price. But yeah, definitely need currency

shawncplus:
Closed with a78e0feacf40abcc915571506c78419a9ab6244d

PR #194, opened 2017-03-10
Shop - initial PR

jackjwilliams:
For #172

### DIFF
--- a/bundles/ranvier-vendors/behaviors/npc/vendor.js
+++ b/bundles/ranvier-vendors/behaviors/npc/vendor.js
@@ -8,13 +8,14 @@ module.exports = (srcPath) => {
     listeners: {
       playerEnter: state => function (config, player) {
         if (config.enterMessage) {
-          B.sayAt(player, config.enterMessage);
+          B.sayAt(player, '');
+          state.ChannelManager.get('say').send(state, this, config.enterMessage);
         }
       },
 
       playerLeave: state => function (config, player) {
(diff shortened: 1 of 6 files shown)


bundles/ranvier-vendors/behaviors/npc/vendor.js
```diff
@@ -0,0 +1,26 @@
...
+
+
+module.exports = (srcPath) => {
+  const Logger = require(srcPath + 'Logger');
+  const B      = require(srcPath + 'Broadcast');
+  return  {
+    listeners: {
+      playerEnter: state => function (config, player) {
+        if (config.enterMessage) {
+          B.sayAt(player, config.enterMessage);
```
shawncplus: I think it'd be cool for NPC messages to the player to use the `say` channel. `state.ChannelManager.get('say').send(state, this, config.enterMessage);` So that coloring/format is consistent between player say/NPC say.

jackjwilliams: Yeah I like that, that way it's more like the NPC is talking directly to the player, instead of spamming the room

shawncplus: kk, makes sense to me

shawncplus:
So I did think of a use for an item behavior for this PR: `sellable` by default items aren't sellable. Sellable would allow you to configure the value of an item.

shawncplus:
Merged manually

PR #201, opened 2017-03-25
Add teleport command to debug-commands bundle

ethryx:
This will demonstrate the ability to teleport to either an online player or a room id. The command works like this:

```
teleport <room/player>
ALIAS: tp
```

### DIFF
--- a/bundles/debug-commands/commands/teleport.js
+++ b/bundles/debug-commands/commands/teleport.js
@@ -4,7 +4,7 @@ module.exports = (srcPath) => {
   const Broadcast = require(srcPath + 'Broadcast');
   const PlayerRoles = require(srcPath + 'PlayerRoles');
   const Player = require(srcPath + 'Player');
-  
+
   return {
     aliases: ['tp'],
     usage: 'teleport <player/room>',
@@ -52,6 +52,10 @@ module.exports = (srcPath) => {
           Broadcast.sayAt(follower, `You stop following ${player.name}.`)
         }
       });
+
+      if (player.isInCombat()) {
+        player.removeFromCombat();
+      }
     }
   }
 };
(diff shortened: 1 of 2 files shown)


samjaninf:
Shouldn't the command be goto? Not that it makes a difference just thought you guys were following common Diku mud commands. Unless goto isn't a common Diku mud thing in which case just ignore me.

ethryx:
That's a good point. I do think `goto` should at the very least be an alias, if not changing the base command name entirely. I'll wait for more input because I'm not sure either.

bundles/debug-commands/commands/teleport.js
```diff
@@ -0,0 +1,61 @@
...
+        Broadcast.sayAt(player, '<b><green>You snap your finger and instantly appear in a new room.</green></b>\r\n');
+        state.CommandManager.get('look').execute('', player);
+      });
+
+      Broadcast.sayAt(oldRoom, `${player.name} teleported away.`);
+      Broadcast.sayAtExcept(targetRoom, `${player.name} teleported here.`, player);
+
+      player.followers.forEach(follower => {
+        follower.unfollow();
+        if (follower instanceof Player) {
```
shawncplus: If you make this `!follower.isNpc` you won't have to include the `Player` class

ethryx: Nice, I just updated the PR to include that change.

bundles/debug-commands/commands/teleport.js
```diff
@@ -0,0 +1,60 @@
...
+      Broadcast.sayAtExcept(targetRoom, `${player.name} teleported here.`, player);
+
+      player.followers.forEach(follower => {
+        follower.unfollow();
+        if (!follower.isNpc) {
+          Broadcast.sayAt(follower, `You stop following ${player.name}.`)
+        }
+      });
+
+      if (player.isInCombat()) {
```
shawncplus: You might want to remove the player from combat/remove their followers before you move them to the next room to prevent the moment where the player is still in combat but not in the room.

ethryx: Good point- I moved that above the actual player movement.

PR #203, opened 2017-03-26
Fix getting the starting room from RoomManager

ethryx:
When a player is hydrated, if the player isn't in a valid room, a call to `state.RoomManager.getStartingRoom()` is made, which is undefined.


shawncplus:
The fix here is to just fix the place that calls `getStartingRoom()` to use `.startingRoom` like https://github.com/shawncplus/ranviermud/blob/staging/bundles/ranvier-input-events/input-events/finish-player.js#L21

ethryx:
Cool, I'll update that.

PR #209, opened 2017-03-31
Check inventory then equipment for containers after checking room.

seanohue:


src/Item.js
```diff
@@ -179,9 +179,13 @@ class Item extends EventEmitter {
 
     // if the item was saved with a custom inventory hydrate it
     if (this.inventory) {
+      // handle equipped containers.
+      if (!(this.inventory instanceof Inventory)) {
```
shawncplus: I'm not entirely sure how this case came up because the constructor does this logic already

seanohue: I thought it did too. It specifically happens when logging back onto a character who is wearing a container that has items in it. For some reason, it was not hydrating property. I should double check and make sure that this behavior isn't a bug specific to my fork :/

shawncplus: Yeah I'm more interested in how this happened than just putting a band-aid on it. If you want to list repro steps I can try to reproduce it when I get home.

seanohue: 1. Add a wearable container to the items.yml file. For instance:
```
- id: 16
  name: "Ragged Canvas Knapsack"
  type: CONTAINER
  roomDesc: "Ragged Canvas Knapsack"
  keywords: [ "rag", "ragged", "knap", "knapsack", "canvas" ]
  description: "A small canvas knapsack attached to a shoulder strap. The strap appears frayed."
  items: [ ]
  maxItems: 3
  slot: 'back'
  quality: common
  properties:
    stats:
      armor: 1
  behaviors:
    sellable:
      value: 2
      currency: gold
```

2. Put said item in a room or container via editing the .yml
3. Log in to a character, acquire the equippable-container-item, put something into it, equip it, save, and log off.
4. Log back in and see if the server crashes immediately due to trying to call hydrate on a plain JS object.

seanohue: If you'd like I can remove this from the PR and test it out to make sure that it's not just an issue on my part.

seanohue: I ended up just removing this part from the PR.

seanohue:
Any other changes need to be made to this? I removed the unneeded fix.

shawncplus:
Nope, just forgot about this PR :)

PR #210, opened 2017-04-16
Make both /r/n and /n line endings compatible with telnet command processor

franciscopreller:
- Have tested change, this makes it compatible to both line endings

### DIFF
--- a/src/Telnet.js
+++ b/src/Telnet.js
@@ -241,7 +241,7 @@ class TelnetStream extends EventEmitter
       this.stream.fresh = false;
       return;
     }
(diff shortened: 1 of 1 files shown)


marcelomoreli:
I had to make this change so that it is compatible with zMud (client)

Now this change should be done on Row #328 on ranvier-telnet/node_modules/ranvier-telnet/index.js

shawncplus:
Hmm okay, I'll have to port this change over to the other lib now

shawncplus:
No longer part of this repo

PR #211, opened 2017-04-16
Feature/add dockerfile option

franciscopreller:
- Adds a simple Dockerfile and .dockerignore for building in Docker
- Docker node version locked down to 7.8
- Node image comes with yarn by default, so we use it for faster building
- Adds yarn.lock file to lock down npm modules by version


franciscopreller:
Sorry, accidentaly included a commit from the other PR I made, will clean this and re-send.

PR #212, opened 2017-04-16
Add Dockerfile option

franciscopreller:
This allows to build Ranvier in Docker.

- Adds a simple Dockerfile and .dockerignore for building in Docker
- Docker node version locked down to 7.8
- Node image comes with yarn by default, so we use it for faster building
- Adds yarn.lock file to lock down npm modules by version


shawncplus:
Probably works but ancient, will reopen later if someone asks for it

PR #213, opened 2017-04-28
Add crit behavior

seanohue:


shawncplus:
Hmm, I was thinking about how to handle "critical" hits and other cases like that: misses, glancing blows, critical hits, dodge, parry, block, etc. I think I'd prefer it somehow integrated directly into Damage itself. Perhaps in the same way that `Damage.evaluate` calls `evaluateOutgoingDamage`/`evaluateIncomignDamage` those methods should handle calculating crit/miss/glancing blows/dodge/parry/block/etc.

seanohue:
That could work -- I was wondering about how to add things like weapon specializations that increase the amount of critical hits... and that may be the best way to do it. 
So, perhaps a crit behavior for a weapon would add an effect on equip that has hooks for the evaluation, and modifies the damage that way?

seanohue:
Made some changes and merged staging back into this branch since things were apparently a bit stale.

Thoughts?

shawncplus:
Personally I wouldn't handle this as an effect/behavior. I'd just give Character a crit chance stat and then have that stat interact inside `evaluateOutgoingDamage`. Then the equipment will just automatically be able to modify stats via `properties.stats` on the definition. No effect/behavior needed and it would also allow skills to do critical damage.

seanohue:
Hmm, I didn't think of that. I guess that would fit most of the edge cases that I thought I was solving by doing it this way, and also be easier to use on the part of the person creating the item. When I have time this week I'll go back and rework it.

The one "edge case" (if it is that) that still stands out is this: we probably want to know whether a hit is "critical" or "glancing" or whatnot _before_ we evaluate outgoing damage. So, where/how should that be done? Right now it's by calling that hook used by the effect. Should it instead just be something like `this.critical = Random.probability(player.getMaxAttribute('criticalChance');` ?

shawncplus:
>  we probably want to know whether a hit is "critical" or "glancing" or whatnot before we evaluate outgoing damage.

There are a couple options.
1. In `Damage.evaluate` in in the `if (this.attacker)` block check against the attacker's crit chance and modify the damage amount before passing it into `Character.evaluateOutgoingDamage`
2. Wait until the damage gets into `Character.evaluateOutgoingDamage` and before it calls `this.effects.evaluateOutgoingDamage` check against the character's crit chance.

It's basically a question of do you want Character to know how to make damage critical, or do you want Damage to know how to make itself critical. Both are viable approaches.

seanohue:
Alright, reworked the implementation. Will do some more testing tomorrow and ping you when it's good to go.

src/EffectList.js
```diff
@@ -166,6 +166,25 @@ class EffectList {
     return Math.max(currentAmount, 0) || 0;
   }
 
+  evaluateCriticalChance(damage) {
```
seanohue: oh no

seanohue:
This should be good to go. 👍

seanohue:
Resolving conflicts on this right quick.

seanohue:
I think this just needs to have the `critical` stat added to characters when they are created. I can likely do that tonight.

seanohue:
Also potentially done?

src/Damage.js
```diff
@@ -25,7 +27,9 @@ class Damage {
       attacker = null,
       type = "physical",
       source = null,
-      hidden = false
+      hidden = false,
+      critical = false,
+      glancing = false
```
shawncplus: I don't see you using `glancing` anywhere so perhaps we can remove that for now.

src/Damage.js
```diff
@@ -51,7 +57,13 @@ class Damage {
    */
   evaluate(target) {
     let amount = this.amount;
+
     if (this.attacker) {
+      const critChance = Math.max(this.attacker.getMaxAttribute('critical') || 0, 0);
```
shawncplus: I'm really really torn on having this part of core. I lean to leaving it in simply because this seems to degrade gracefully, insofar as if the character doesn't have the critical attribute it just never does a critical hit.

src/Damage.js
```diff
@@ -51,7 +57,13 @@ class Damage {
...
+      this.critical = Random.probability(critChance);
+      if (this.critical) {
+        amount = Math.ceil(amount * 1.5);
```
shawncplus: I think this amount multiplier should be a configuration option on the Damage constructor that way the skill/effect/whatever can control the amount. Example: a fireball does 4x crit damage or whatever.

seanohue:
Updorted.

PR #217, opened 2017-05-21
Doc edits

seanohue:
Did a once-over of the docs to fill in some knowledge gaps and prepare for some Ranvier evangelism. Noticed some small issues and just decided to fix them as I went.
Some of it is a bit nitpicky, but I tried to focus on adding clarity to certain parts rather than just adding punctuation or correcting grammar/formatting.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -2,12 +2,12 @@
 
 # ranvier
 
-Node.js based MUD engine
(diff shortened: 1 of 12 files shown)


shawncplus:
This wasn't a small amount of work, great job.

PR #218, opened 2017-05-21
Multi resource costs

seanohue:
Now the resource costs can be an array or a single object, as long as each object has an attribute and a cost property.

### DIFF
--- a/bundles/ranvier-commands/commands/put.js
+++ b/bundles/ranvier-commands/commands/put.js
@@ -25,7 +25,9 @@ module.exports = (srcPath) => {
       const fromArg = parts[0];
       const toArg = parts[1];
       const item = Parser.parseDot(fromArg, fromList);
-      const toContainer = Parser.parseDot(toArg, player.room.items);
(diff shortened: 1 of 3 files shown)


bundles/ranvier-commands/commands/put.js
```diff
@@ -25,7 +25,9 @@ module.exports = (srcPath) => {
       const fromArg = parts[0];
       const toArg = parts[1];
       const item = Parser.parseDot(fromArg, fromList);
-      const toContainer = Parser.parseDot(toArg, player.room.items);
+      const toContainer = Parser.parseDot(toArg, player.room.items) ||
+                          Parser.parseDot(toArg, player.inventory) ||
+                          Parser.parseDot(toArg, player.equipment);
```
seanohue: Woops, this commit snuck in. I'll chop it out.

Should also probably update the docs.

seanohue: Alright, removed the irrelevant commit and added some documentation.

shawncplus:
LGTM

PR #221, opened 2017-06-23
Round overall goal percent that is displayed to avoid long repeating …

seanohue:
…numbers.


seanohue:
As an aside, it may be time for a 1.0.2 patch release?

shawncplus:
Yeah, I was thinking that too

PR #226, opened 2017-07-12
Bugfix for "You followed <player> to undefined." when you're following or in a group with someone.

RCopeland:


shawncplus:
Ooh, nice catch.

PR #228, opened 2017-07-16
Extract Data.isScriptFile method

seanohue:
Small refactor.

This code was repeated a fair amount, and I can think of instances where a bundle may want to use this method when loading something dynamically.

I decided to move it to the Data class since it's a pure-ish function and not exclusively related to loading a bundle, so thought it was a good candidate for a static utility method.


shawncplus:
Merged

PR #230, opened 2017-07-17
Use for/of rather than combining for/in with a hasOwnProperties check.

seanohue:
For readability and best practices.

### DIFF
--- a/data/player/Admin.json
+++ b/data/player/Admin.json
@@ -31,21 +31,77 @@
   },
   "level": 1,
   "name": "Admin",
-  "room": "limbo:1",
-  "effects": [],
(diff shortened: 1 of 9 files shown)


data/player/Admin.json
```diff
@@ -31,22 +31,78 @@
   },
   "level": 1,
   "name": "Admin",
-  "room": "limbo:1",
-  "effects": [],
+  "room": "limbo:8",
```
seanohue: why am i like this

shawncplus: HA

seanohue: Reverted 💥

PR #232, opened 2017-07-25
Fix Multiplay Bug

### LINKED ISSUE #206: Bug with multiplay on the same character
### Description
It's possible to login to the same character with neither socket being disconnected.

### Expected outcome
When I login to a character that is already logged in, with multiplay enabled, the existing instance should be terminated.

### Actual outcome
When I login to a character that is already logged in, with multiplay enabled, both connections remain connected.

### Steps to reproduce
1. Login to a character
2. In another instance, login to the same character. The original connection will be disconnected, as expected
3. In yet another instance, login to the same character again. Both the new and old connections will remain connected.

### Version of Ranvier
1.0.1

seanohue:
Introduce an async delay so that events happen and the player is consistently booted.

Fixes #206

### DIFF
--- a/bundles/ranvier-input-events/input-events/choose-character.js
+++ b/bundles/ranvier-input-events/input-events/choose-character.js
@@ -15,6 +15,10 @@ module.exports = (srcPath) => {
 
       const say = EventUtil.genSay(socket);
 
+      const delay = (time = 0) => (new Promise(resolve => {
+        setTimeout(resolve, time);
+      }));
+
       /*
       Player selection menu:
       * Can select existing player
@@ -29,7 +33,6 @@ module.exports = (srcPath) => {
       const maxCharacters   = Config.get("maxCharacters");
       const canAddCharacter = characters.length < maxCharacters;
       const canMultiplay    = Config.get("allowMultiplay");
-
 
       let options = [];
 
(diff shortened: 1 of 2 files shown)


shawncplus:
lol nooooo, delays are the wrong way to fix race conditions

seanohue:
But they are still _a_ way to fix a race condition...right?

I did find something interesting prior to introducing the delay. After logging into an already logged-in character for the second time, there were no players in the PlayerManager set when `handleMultiplaying` got called. 

That said, I don't know how to fix that right now so I'll come back to this in the morning.

seanohue:
Ended up playing around with async/await and promises to try to fix this and lo and behold, it works! This simply does not resolve the async computation until the player's socket emits the close event.

shawncplus:
This is a much better solution, will try this out later

shawncplus:
Does async/await work with version 7? If not we'll have to bump the requirements up from 7+

seanohue:
It does not. We can bump the requirements up to 8 if needed, or put instructions for running this with a harmony flag for later versions of 7, or I can rewrite this to simply use native promises (which wouldn't be difficult, I mainly wanted to learn how to use `async` and `await` on something practical 💯  )

seanohue:
Thoughts on using async/await vs. promises?

seanohue:
Alright, re-implemented using Promises. Did the same tests as before and it seems to work fine. Still fairly easy to follow in terms of code and compatible with Node 7 and 8.

PR #235, opened 2017-07-28
Bug report

seanohue:
Tried to follow best practices and leave the implementation fairly basic. It is in its own bundle that can be easily swapped out for another, and leverages the event system (so it may be possible to use the default bug reporting and add an additional bundle that _also_, say, automatically emails the server admin).

It can also broadcast to any online admins (or builders in the case of a non-bug), if configured to do so.

### DIFF
--- a/src/ChannelAudience/RoleAudience.js
+++ b/src/ChannelAudience/RoleAudience.js
@@ -10,7 +10,7 @@ class RoleAudience extends ChannelAudience {
   }
 
   getBroadcastTargets() {
-    return [...this.state.PlayerManager.players].filter(player => player.role >= this.minRole);
+    return this.state.PlayerManager.filter(player => player.role >= this.minRole);
   }
 }
 
--- a/src/ChannelAudience/RoleAudience.js
+++ b/src/ChannelAudience/RoleAudience.js
@@ -1,0 +1,15 @@
+class RoleAudience extends ChannelAudience {
+  constructor(options) {
+    super(options);
+    if (!options.hasOwnProperty('minRole')) {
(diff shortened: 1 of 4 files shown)


shawncplus:
Nice usage of `arg0` :+1:

shawncplus:
The only suggestion I would have is that I was talking to @criticalfault yesterday about an `AdminAudience` for broadcast so it might be useful to create one instead of manually looping over `state.PlayerManager.players`

e.g.,

    class RoleAudience extends ChannelAudience {
      constructor(options) {
        super(options);
        if (!options.hasOwnProperty('minRole')) {
          throw new Error('No role given for role audience');
        }
        this.minRole = options.minRole;
      }
      
      getBroadcastTargets() {
        return this.state.PlayerManager.players.filter(player => player.role >= this.minRole);
      }
    }

`Broadcast.sayAt(new AdminAudience(), 'some message')`

seanohue:
That's a solid addition to core. I'll make that change.

src/ChannelAudience/RoleAudience.js
```diff
@@ -0,0 +1,17 @@
...
+  constructor(options) {
+    super(options);
+    if (!options.hasOwnProperty('minRole')) {
+      throw new Error('No role given for role audience');
+    }
+    this.minRole = options.minRole;
+  }
+
+  getBroadcastTargets() {
+    return [...this.state.PlayerManager.players].filter(player => player.role >= this.minRole);
```
shawncplus: looking at the other channel audience this line can just be `this.state.PlayerManager.filter`  instead of directly accessing `PlayerManager.player`

Issue #239, opened 2017-08-01
Inaccurate helpfile for default craft command


seanohue:
### Description
When typing `help craft` the helpfile appears to contain a copy+paste error.
### Expected outcome
The `help craft` output should be specifically tailored to the `craft` command.

### Actual outcome
```
--------------------------------------------------------------------------------
                                     craft                                      
--------------------------------------------------------------------------------
Syntax: quest <log/list/complete/start> [npc] [number]

craft list                  View crafting recipe categories
craft list <category #>     List recipes for a given category
craft list   <cat #> <item #> View stats and ingredients for a given item
craft create <cat #> <item #> Create item viewed with list <#> <#>
--------------------------------------------------------------------------------
```

### Steps to reproduce

On the `staging` branch, pull down the latest version, boot it up, log in, and type `help craft`.

### Version of Ranvier
v1.1.1

shawncplus:
Closed by #242

PR #242, opened 2017-08-01
Fix craft helpfile

seanohue:
Also did some refactoring along the way with the help command and rendering of helpfiles/error handling.

### DIFF
--- a/bundles/ranvier-commands/commands/help.js
+++ b/bundles/ranvier-commands/commands/help.js
@@ -24,8 +24,13 @@ module.exports = (srcPath) => {
         Logger.error(`MISSING-HELP: [${args}]`);
         return B.sayAt(player, "Sorry, I couldn't find an entry for that topic.");
       }
-
-      B.sayAt(player, render(state, hfile));
+      try {
+        B.sayAt(player, render(state, hfile));
+      } catch (e) {
+        Logger.warn(`UNRENDERABLE-HELP: [${args}]`);
+        Logger.warn(e);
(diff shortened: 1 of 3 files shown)


seanohue:
Resolves #239

bundles/ranvier-crafting/help/craft.yml
```diff
@@ -1,5 +1,5 @@
 ---
-command: quest
+command: craft <list/create> [category #] [item #]
```
shawncplus: This is the wrong way to do it, the purpose of the `command` entry in the helpfile is to do a `CommandManager` lookup to get the `Command.usage` property. So this should not be necessary

bundles/ranvier-commands/commands/help.js
```diff
@@ -40,7 +45,7 @@ module.exports = (srcPath) => {
 
     const formatHeaderItem = (item, value) => `${item}: ${value}\r\n\r\n`;
     if (hfile.command) {
-      let actualCommand = state.CommandManager.get(hfile.command);
+      let actualCommand = state.CommandManager.get(hfile.command) || state.CommandManager.get(hfile.command.split(' ')[0]);
```
shawncplus: See comment below about `Command.usage`

bundles/ranvier-crafting/help/craft.yml
seanohue: Fixed.

shawncplus:
:+1:

PR #244, opened 2017-08-01
Load npc equipment from file

pointybird:
Change to load npc equipped items from yml file similar to way players do.

### DIFF
--- a/src/Npc.js
+++ b/src/Npc.js
@@ -26,7 +26,7 @@ class Npc extends Character {
     this.area = data.area;
     this.behaviors = new Map(Object.entries(data.behaviors || {}));
     this.damage = data.damage;
-    this.defaultEquipment = data.equipment || [];
+    this.defaultEquipment = data.equipment || new Map();
     this.defaultItems = data.items || [];
     this.description = data.description;
     this.entityReference = data.entityReference; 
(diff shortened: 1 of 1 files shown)


pointybird:
After thinking about it some more, I think this should probably change a bit. I was changing the slot on load to handle where the file wants it, specifically to handle cases like dual wielding. Equip only goes off item.slot which is just a simple string, so if you had 2 daggers with slot = 'wield', the first would get equipped, the second would throw an exception when wielding since the slot is taken. A better way would probably be to let equip take an optional second parameter specifying where to wield, and then defaulting to item.slot if it's not given, so we could call equip(dagger) for the first one and equip(dagger, "offhand") for the second. Loading from file would then leave item.slot untouched but pass in the location to the call to equip.

seanohue:
I think that could be a useful feature. Perhaps could also allow for item.slot to be an array of strings in the order of preference when equipping.

seanohue:
Added a PR to enable multiple slots. Perhaps once that is merged, this PR could be edited to leverage that feature?

pointybird:
Sounds good

shawncplus:
This PR is pretty stale since it was made before the change to how `equip()` works. @pointybird I'm going to close this for the moment just to clean up the PR queue but I'm more than willing to merge this in if you resubmit

PR #245, opened 2017-08-02
Added extra descriptions.

thedandroid:
Extra descriptions may be added to keywords in the room description.
Characters may now use look [keyword] for closer inspection.

### DIFF
--- a/src/Room.js
+++ b/src/Room.js
@@ -17,6 +17,7 @@ const Logger = require('./Logger');
  * @property {string}        script       Name of custom script attached to this room
  * @property {string}        title        Title shown on look/scan
  * @property {object}        doors        Doors restricting access to this room. See documentation for format
+ * @property {Array<object>} exdescs      Extra descriptions in the room { id: string, desc: string }
  * @extends EventEmitter
  * @listens Room#updateTick
  */
@@ -43,6 +44,7 @@ class Room extends EventEmitter {
     // create by-val copies of the doors config so the lock/unlock don't accidentally modify the original definition
     this.doors = new Map(Object.entries(JSON.parse(JSON.stringify(def.doors || {}))));
     this.defaultDoors = def.doors;
+    this.exdescs = def.exdescs || [];
 
     this.items = new Set();
     this.npcs = new Set();
--- a/src/Room.js
+++ b/src/Room.js
@@ -17,7 +17,7 @@ const Logger = require('./Logger');
  * @property {string}        script       Name of custom script attached to this room
  * @property {string}        title        Title shown on look/scan
  * @property {object}        doors        Doors restricting access to this room. See documentation for format
- * @property {Array<object>} exdescs      Extra descriptions in the room { id: string, desc: string }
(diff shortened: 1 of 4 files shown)


thedandroid:
Go easy on me, I'm still learning! 😅

seanohue:
So, I like this as a concept and don't see any issues with the code (without having pulled it down and tested it yet). One thing I'd change is `exdesc` to `extraDesc` or `detailDesc` simply because I'm not a fan of unclear abbreviations. OTOH maybe `exdesc` is a common nomenclature for this kind of thing in MUDs and I just don't know it.

I also think the method for finding the description might be better as a part of the `Room` class rather than `RoomManager`.

I can see this being super handy in a text adventure type game, tbh.

thedandroid:
I totally agree with the name change. It came from Smaug, where I got the concept, and just never thought past the name once I started working on it.

Moved to the Room class.

They were super useful when I was building, to not clutter up a room description with minutiae of some object in the room that wasn't a physical object.

seanohue:
Agreed, they could be handy for adding little details. I opted to go the route of making decor out of `Item`s with `noPickup: true`, with the bonus of being able to script them and add behavior as needed. But having this as an option is cool. I also like the text highlighting feature.

src/Room.js
```diff
@@ -258,6 +258,25 @@ class Room extends EventEmitter {
     door.locked = true;
   }
 
+/**
+  * Get the detail descriptions contained in a room by id
+  * @param {string} id    detail description id
+  * @return {false|Object}
+  */
+  findDetailDesc(detailDescId) {
+    const detailDescs = Array.from(this.detailDescs).filter(dd => dd.id.indexOf(detailDescId) === 0);
```
seanohue: This could use the `String` method `startsWith` rather than `indexOf`.

src/Room.js
```diff
@@ -258,6 +258,25 @@ class Room extends EventEmitter {
...
+
+    if (!detailDescs.length) {
```
seanohue: Could these two conditionals be condensed into a single one such as `if (detailDescs.length !== 1)`?

shawncplus:
I don't think this is a bad feature but I'm going to have to think about it because `findDetailDescs` is a very opinionated function to add to src as it would serve absolutely no purpose if that specific implementation of the `look` command wasn't used.

seanohue:
I really like the idea of "disambiguation of conflict commands", I think it could be implemented by allowing bundles to be weighted:

(ranvier.json)
```json
"bundles": {
  "ranvier-commands": 1
  "custom-commands": 2
}
```
In this case, any commands from `custom-commands` that are also in `ranvier-commands` would clobber the ranvier default versions.

As far as implementing detailDescs without adding an opinionated method to `src` -- I think that this same functionality could be accomplished by allowing rooms to have a key/value store of properties defined in .yaml, as is the case with items.

shawncplus:
I'm going to close this PR as I'm not comfortable merging the core change given that we came up with the idea of command disambiguation which I would call a blocker for this.

Issue #247, opened 2017-08-02
Items are searched by get command in confusing order.


seanohue:
### Description

When looting corpses, for example, one might expect the one that was felled most recently to be the first one searched (since it is least likely to be empty). Since room items are Sets, and Sets are iterated over in insertion order, it may be useful to have a method for iterating in reverse order. (easiest way is just to convert the Set to an Array then reverse the Array, inside of the `get` command script).

### Expected outcome
If I kill a mob, loot its corpse, then kill a second in the same room and loot its corpse, the `loot` or `get all` command will search the most recently created corpse container. (gruesome, I know)

### Actual outcome
The `get all` command searches the first-inserted corpse container, which is the empty one, so I have to do `get all 2.corpse`.

### Steps to reproduce

Log onto the default server, move to a room with multiple training dummies, and kill one of them. Loot it, kill another, and attempt to loot the second without using dot notation.

### Version of Ranvier
v1.1.1

seanohue:
Updated this since the solution of abstracting a RoomEntitiesList seems silly to fix just a single edge case.

PR #251, opened 2017-08-05
Fix the get command, closes #240

### LINKED ISSUE #240: 'get all' or 'loot' on a closed chest shows one failure message per item in the container
### Description
When trying to get all from a closed container, one sees one message per item in the container, explaining that the container is closed.

### Expected outcome
When interacting with a closed container in a way that would require it to be open, I would expect a single message telling me that I cannot look inside the container because it was closed.

### Actual outcome
I receive a single identical message per item in the container, telling me that the container is closed.

### Steps to reproduce
Boot up the staging server, log in and try to `get all chest` in the first room.

### Version of Ranvier
`v1.1.1`

miki123211:
When trying to get all items from a closed container, only one failure message is now displayed.

### DIFF
--- a/bundles/ranvier-commands/commands/get.js
+++ b/bundles/ranvier-commands/commands/get.js
@@ -45,8 +45,12 @@ module.exports = (srcPath) => {
         }
 
         if (container.type !== ItemType.CONTAINER) {
-          return Broadcast.sayAt(player, `${container.name} isn't a container.`);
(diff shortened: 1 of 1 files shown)


miki123211:
Added another commit to also fix #247, because those were kind of related so I decided to fix both in one go

shawncplus:
LGTM

PR #253, opened 2017-08-05
Fix craft list display.

thedandroid:


shawncplus:
This doesn't look quite right but I may be wrong, will try it out shortly

thedandroid:
Thought the same thing later after I was already away from my desk. I’ll test later when I get home.

thedandroid:
Unless you beat me to it.

shawncplus:
This is _technically_ correct but I'd rename the `items` variable to something like `categoryEntry`.

PR #254, opened 2017-08-08
Multiple item slots

### LINKED ISSUE #249: Allow items to be equipped in multiple slots.
### Description

Example: You want to implement the ability to hold two items at once, one "wielded" and one "offhand". Currently, the `Item` class only allows for a single `slot` to be described, as a string. So if I have a candlestick that I equip and it goes into the 'wielded' slot, and then I try to equip a pitchfork, I will receive an error message telling me I cannot hold both. One potential solution is to allow item.slot to be defined as either a single slot (a string) or multiple slots (an array of strings). If it has multiple defined slots, then equip can do one of two things:
1) Try each of the slots in order until it finds an empty one, or throw an error if it does not find an open slot.
2) Take an optional second argument `equip pitchfork offhand` to specify an equipment slot. If the specified slot is not included in the pitchfork's array of slots (such as `equip pitchfork sandwich`), then throw an error, otherwise try to equip the item in that slot (succeeding or failing as normal).

### Expected outcome

Items can have multiple equippable slots.
Trying to equip an item with multiple equippable slots will try the slots in order until finding an open one or failing.
You can specify a slot to attempt to equip the item in.

### Actual outcome

Items only have one slot and there can only be one equipped item per slot.

### Steps to reproduce

Try creating an item with multiple slots. You can't do it. I double-dare you.

### Version of Ranvier
v1.1.1

seanohue:
Resolves #249 .

The `slot` property in each items YAML definition can now be a string or array of strings.
The appropriate `EquipError` is thrown if all of the listed slots are occupied by equipped items.

### DIFF
--- a/bundles/ranvier-areas/areas/limbo/items.yml
+++ b/bundles/ranvier-areas/areas/limbo/items.yml
@@ -27,9 +27,9 @@
   roomDesc: "A wooden chest rests in the corner, its hinges badly rusted."
   keywords: [ "wooden", "chest" ]
   description: "Time has not been kind to this chest. It seems to be held together solely by the dirt and rust."
-  items: [ "limbo:1", "limbo:6", "limbo:7", "limbo:8", "limbo:leather_gauntlet" ]
+  items: [ "limbo:1", "limbo:6", "limbo:7", "limbo:8", "limbo:leather_gauntlet", "limbo:leather_gauntlet" ]
   closed: true
-  maxItems: 5
+  maxItems: 6
   properties:
     noPickup: true
 - id: 4
@@ -59,7 +59,7 @@
     stats:
       strength: 2
       stamina: 2
-  behaviors: 
+  behaviors:
     sellable:
       value: 30
       currency: gold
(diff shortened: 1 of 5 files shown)


shawncplus:
This LGTM at the moment, were you still working on this or is it ready to go as is?

seanohue:
Yes, though as discussed in the slack, this only implements an either/or relationship between the list of slots. So, I can have an individual item that is wearable in two slots, but i can't make an item, for example, `pair of magical gauntlets` that covers both the `left hand` and `right hand` slots.

shawncplus:
I think the unopinionated approach is to make the equip command _require_ the slot argument. That way the `equip` command can be the one that's opinionated about what happens

seanohue:
Agreed, will rework this to have the core be largely un-opinionated.

marcelomoreli:
Is this still on-going?

shawncplus:
I think I'm going to close this PR, the conclusion that was arrived at was that the core `equip()` command shouldn't look at `item.slot` directly, instead it should take the slot as an argument and the `slot` property should be removed from item. That way it's up to the command to decide what to pass `equip()` and how.

PR #255, opened 2017-08-09
Kick when idle

### LINKED ISSUE #204: Kick Idle Players
### Description

As an admin, I'd like to be able to configure a maximum idle time and kick players above that threshold.
If there is no idle time set, players will not be kicked.
This might require a few changes to see when the player last entered a command, but I don't think it would require any major refactors.

### Expected outcome
Idling forever can lead to being kicked.

### Actual outcome
I can idle infinitely without being kicked and there is no way to configure otherwise.

### Steps to reproduce
Idle forever.

### Version of Ranvier
1.0.1

seanohue:
Resolves #204


shawncplus:
Needs conflict resolution

seanohue:
Resolved conflicts.

shawncplus:
I don't think you have to use meta for this since there's no real value in persisting it. Just set it right on the player `player._lastCommandTime`, no need to create a property in the core Player class or anything.

shawncplus:
@seanohue just a ping to remind you of the last comment there

seanohue:
doin' th' thing

seanohue:
Done!

PR #263, opened 2017-08-23
Added aliases to channels - and to the bundle as example

marcelomoreli:
Adding aliases to channels. Added in '.' as alias to the chat as example in the bundle.


shawncplus:
I would appreciate if you also updated the channel docs to say that this feature exists, the file is `docs/extending/channels.md`

marcelomoreli:
I'm not 100% sure on how to test the documentation, but It's done!

Issue #163, opened 2017-03-03
Emoting


seanohue:
I would expect emoting to, at a minimum, allow me to do things like this:

`> emote dances a jig.`

And I would see:
`>You emote "Sean dances a jig."`

And others would see:
`>Sean dances a jig.`

To extend this, it would be nice to have shorthand (`:` is conventional, I think), and potentially to be able to refer to specific characters or entities using `~`. However, these features might be something for post-1.0 or for individual bundle developers to figure out.

marado:
I'm not sure there's a need for the feedback to the user to be distinct from the output to all others in the room. ie, mostly I see emote implemented as:
```
emote dances a jig.
Sean dances a jig.
```

I agree that `:` is the most popular shortcut to the emote command.

seanohue:
I agree with the exception that your emote targets a specific entity (player, NPC) in the room.

However for an MVP it is probably fine to exclude that.

shawncplus:
Closed with #268

PR #264, opened 2017-08-28
Add the "emote" command (#163)

brandonshults:
From the player's perspective:
image

From a bystander's perspective:
image


shawncplus:
Might be useful to use `CommandParser.parseDot` instead of the custom `findTargetByName` that way if there's more than one 'chest' item in the room it can be targeted.

brandonshults:
I've now improved the PR to use ```CommandParser.parseDot.```

PR #265, opened 2017-08-30
Fix for issue with recall: no energy attribute is causing issues

jackjwilliams:
Fix issues with VSCode launch.json config
Fix issue with recall.js checking the energy attribute (it's never created when you create a new character), throwing an error.

To reproduce the error: create a new paladin, waypoint save, waypoint home, then try to recall.

### DIFF
--- a/bundles/ranvier-commands/commands/recall.js
+++ b/bundles/ranvier-commands/commands/recall.js
@@ -11,13 +11,6 @@ module.exports = srcPath => {
         return B.sayAt(player, 'You do not have a home waypoint set.');
       }
 
-      if (player.getAttribute('energy') < (player.getMaxAttribute('energy') / 3)) {
-        return B.sayAt(player, 'You do not have enough energy to recall.');
-      }
-
-      const cost = Math.round(player.getMaxAttribute('energy') / 3);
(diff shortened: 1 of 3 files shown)


shawncplus:
I think the better solution here would just be to remove the energy requirement from recall/waypoint usage.

jackjwilliams:
So just remove the checks and allow recalling anytime?

shawncplus:
yeah, for recall and waypoints which also cost energy I believe

bundles/ranvier-commands/commands/recall.js
```diff
@@ -11,13 +11,6 @@ module.exports = srcPath => {
         return B.sayAt(player, 'You do not have a home waypoint set.');
       }
 
-      if (player.getAttribute('energy') < (player.getMaxAttribute('energy') / 3)) {
-        return B.sayAt(player, 'You do not have enough energy to recall.');
-      }
-
-      const cost = Math.round(player.getMaxAttribute('energy') / 3);
-      player.lowerAttribute('energy', cost);
```
shawncplus: ha, good thing you're removing this. This was _completely_ wrong in the first place. `lowerAttribute` is a crazy dangerous method to use.

PR #274, opened 2017-09-05
Feature/web api counts

jackjwilliams:
Note that I realize a Web API refactor to bundles is pending, but the code in the Api area I believe is still valid (IE it will just be moved to a bundle).

Added counts for initial rBuilder Dashboard at https://github.com/jackjwilliams/rbuilder

Added CORS to allow requests to come from outside of localhost:9001 (my ranvier builder runs on localhost:3000 for now)


shawncplus:
What about the classes in `DataManagement` as far as I know those were added for the web api and are only used there.

jackjwilliams:
That's true, they can be moved to the bundle as well. I'll do that next.

jackjwilliams:
I added a log stream for Winston, to WSS. So we can do stuff like this. I'm wondering if I can reverse the dependency somehow to take it out of the core dependencies?

bundles/ranvier-webapi/server-events/webapi.js
```diff
@@ -0,0 +1,66 @@
...
+            }
+          });
+        };
+
+        const init = () => {
+          Logger.log('[WEB] Initializing...');
+          const port = Config.get('webPort') || 9000;
+          setupMiddleWare();
+          const server = http.createServer(app);
+          Logger.getWinston().add(winstonWS.WSTransport, { wsoptions: { server: server, path: '/logs' }});
```
seanohue: Cool -- will this allow us to view Winston logs using the web API?

jackjwilliams: Yeah, the screenshot above shows it in action. ws://localhost:9001/logs is the endpoint

seanohue:
Ah, I see now. What do you mean by "reverse the dependency"?

jackjwilliams:
I didn't want the dependency in core, just in my bundles packages.config

seanohue:
Should be doable. Just put it in a `package.json` file at the bundle's root then be sure to run the bundle-install command.

jackjwilliams:
I fixed those conflicts, and moved the dependency out. Was there a reason this hasn't been merged? @shawncplus

shawncplus:
I'm still seeing conflicts, did you not push your branch?

seanohue:
I'm seeing them, too:
```
Conflicting files
ranvier.json
src/Data.js
```

jackjwilliams:
The data.json can probably just be reverted, the ranvier.json was probably me just enabling the new web app bundle. I'll take a look next chance I get.

PR #283, opened 2017-09-22
Added mobManager and added some validations

marcelomoreli:
Added mobManager so that we can create skill/spells/commands that searches for a mob anywhere in the world, same we have for item and player

### DIFF
--- a/bundles/ranvier-combat/lib/Combat.js
+++ b/bundles/ranvier-combat/lib/Combat.js
@@ -19,7 +19,7 @@ class Combat {
    */
   static updateRound(state, attacker) {
     if (attacker.getAttribute('health') <= 0) {
-      Combat.handleDeath(attacker);
+      Combat.handleDeath(state, attacker);
       return false;
     }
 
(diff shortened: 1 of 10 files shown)


src/Character.js
```diff
@@ -110,7 +110,9 @@ class Character extends EventEmitter
    */
   getMaxAttribute(attr) {
     const attribute = this.attributes.get(attr);
-    return this.effects.evaluateAttribute(attribute);
+    if (this.hasAttribute(attr))
```
shawncplus: this has check should happen before the get

bundles/ranvier-combat/lib/Combat.js
```diff
@@ -19,7 +19,7 @@ class Combat {
    */
   static updateRound(state, attacker) {
     if (attacker.getAttribute('health') <= 0) {
-      Combat.handleDeath(attacker);
+      Combat.handleDeath(state, attacker);
```
shawncplus: Something about having to pass the state around everywhere here feels wrong, I'll take a longer look tomorrow afternoon

marcelomoreli: Yeah, I agree with you, it is strange to pass it everywhere; Let me know if there is a better way to do that

PR #286, opened 2017-09-23
Delete char and account functionality

marcelomoreli:
Added delete char and account functionality and metadata to Account


shawncplus:
This is a good PR, really appreaciate it

PR #287, opened 2017-09-24
MobManager

marcelomoreli:
Added MobManager and some little validations and fixes

### DIFF
--- a/src/Character.js
+++ b/src/Character.js
@@ -109,8 +109,11 @@ class Character extends EventEmitter
    * @return {number}
    */
   getMaxAttribute(attr) {
-    const attribute = this.attributes.get(attr);
-    return this.effects.evaluateAttribute(attribute);
+    if (this.hasAttribute(attr)) {
+      const attribute = this.attributes.get(attr);
+      return this.effects.evaluateAttribute(attribute);
+    }
+    return null;
   }
 
   /**
@@ -126,7 +129,10 @@ class Character extends EventEmitter
    * @return {number}
   */
   getAttribute(attr) {
-    return this.getMaxAttribute(attr) + this.attributes.get(attr).delta;
+    if (this.hasAttribute(attr)) {
+      return this.getMaxAttribute(attr) + this.attributes.get(attr).delta;
+    }
+    return null;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/MobManager.js
```diff
@@ -0,0 +1,34 @@
...
+  constructor() {
+    this.mobs = new Map();
+  }
+
+  /**
+   * @param {string} entityRef
+   * @return {Mob}
+   */
+  getMob(entityRef) {
+    return this.mobs.get(mob.uuid);
```
shawncplus: I think this is probably a copy/paste error. This should probably be iterating over the mobs and finding by entityRef. That said what is this getter used for? Should it get all mobs matching that entity ref? `ItemManager` doesn't have a get by entityRef mainly because such a method would bring up some considerations that need to be made about storage.

MobManager/ItemManager could potentially be some very large lists as such doing a linear search through them probably isn't the best, as such some optimizations could be made such as maintaining two internal maps. One which are UUID to Npc instance, another being a map of EntityRef to array of UUIDs currently in the list matching this ref. That turns the getter from an O(n) search to an O(1), much better.

marcelomoreli: Actually I don't even need the getter, I'm most interested on the setter and having the list available to iterate when needed, with specific conditions.
Should I just remove the getter?

shawncplus: I think it's easier to remove it now and add it back later if we need it

marcelomoreli: Done!

marcelomoreli:
Is this ready to go? I'm expecting this to continue some work from my side :)

src/Character.js
```diff
@@ -109,8 +109,11 @@ class Character extends EventEmitter
    * @return {number}
    */
   getMaxAttribute(attr) {
-    const attribute = this.attributes.get(attr);
-    return this.effects.evaluateAttribute(attribute);
+    if (this.hasAttribute(attr)) {
+      const attribute = this.attributes.get(attr);
```
shawncplus: You don't have to change this but just for future reference my preference here is "return early" style. Meaning you would invert the condition and `return null` inside the if. Then do the real logic outside of the if.

shawncplus:
Looks like it'd work well enough. Might run into weirdness if NPCs are every removed from the game in ways besides death but I think we can deal with that when it comes up

PR #290, opened 2017-09-29
Fix for handleMultiplaying

marcelomoreli:
As pointed by @vehn, its a string, not an object - tested and the fix works great.


marcelomoreli:
Wait - there is a different fix for that

marcelomoreli:
Done, tested with allowMultiplayer false and true.

PR #303, opened 2017-11-18
Add music and more sfx

seanohue:
This is the audio PR plus implementation of menu music and sound effects for each skill.


seanohue:
Should this all be documented as well? Is there a section in the docs specifically for Neuro/WebSockets features? Should there be? Should it be in this repo or the Neuro one? :thinking:

shawncplus:
There is no place to document those things, essentially by design. I don't want Ranvier to require Neuro to be useful and vice versa. I think if they were documented it should be within the README of Neuro and not in the official Ranvier docs for the reason that Ranvier doesn't actually care what features Neuro has since all that stuff is within bundles and we don't document specific bundle features.

shawncplus:
Status on this one? I can't remember if this was good to go or not

seanohue:
Good to go. It needs a merge in Neuro as well (obviously) and perhaps more documentation on the Neuro end of things.

PR #306, opened 2017-12-17
Fix issue with only lowercase player name working with PvP combat (and a BountyGoal.js)

bostrt:


bostrt:
I'm sorry, I did not mean to include the `BountyGoal.js` in this PR but it is present. If you are sensitive about the github audit trail, please let me know and I'll refile this PR.

shawncplus:
Nah, it's fine. Are you still working on the bounty goal or is it done? If you're still working on it I'll wait to merge

bundles/ranvier-quests/lib/BountyGoal.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+const QuestGoal = require('../../../src/QuestGoal');
+
+/**
+ * A quest goal requiring the player to locate and/or return an NPC.
```
seanohue: This is super cool even if it doesn't address the original issue ;) I think it could be a worthy contribution to the core library.

bundles/ranvier-quests/lib/BountyGoal.js
```diff
@@ -0,0 +1,74 @@
...
+    if (this.state.found) {
+      if (room.entityReference == this.config.home) {
+        // Check if we have taken the NPC home
+        this.state.delivered = true;
+      }
+      this.emit('progress', this.getProgress());      
+    } else {
+      let located = false;
+      const goalnpcid = this.config.npc;
+      if (goalnpcid != null) {
```
seanohue: Should there be an error if goalNpcId is null? Also, I would suggest using `goalNpcId !== null` for strict equality... slightly pedantic but could prevent weird bugs that might otherwise be hard to track down.

shawncplus: yeah I think what I could do is making `QuestGoal` have a static `validate()` method that is called on bootup so the server won't start if you have invalid quest configurations. Like this goal could validate that it has a goal NPC

bostrt: @seanohue You are correct, there's a problem if the `goalnpcid` is null. I'll update that with a `!== null` check until @shawncplus can work in a `validate()` method.

PR #313, opened 2018-01-05
correct a few typos in docs

montsamu:


docs/extending/bundles.md
```diff
@@ -55,25 +55,25 @@ A bundle can contain any or all of the following children though it's suggested
 possible. Click on any of the items below to see an in-depth tutorial.
 
 <pre>
-<a href="../areas/">areas/</a>
+<a href="./areas/index.md">areas/</a>
```
shawncplus: My guess is that you have a too-recent version of mkdocs. readthedocs (the doc hosting we use) uses mkdocs version 0.15.0.

As such the only link that needs to be updated for this PR is the `../events/` link in `extending/bundles.md` should be updated to say `../input_events/`

PR #319, opened 2018-02-10
Quest require dupe

seanohue:
Decided to take a shot at this to get a better understanding of how the quest stuff works now. Let me know if I missed anything. Tested it with the intro area quests.


shawncplus:
I'm going to close this because I want to spend some more time thinking about the quest requirements work and it might cause this to conflict.

PR #321, opened 2018-03-27
Update README.mkd

MarchosiasM:
Hey, I just changed the windows install command in the readme file

PR #326, opened 2018-04-06
Added one node-script for all platforms

nahilep:
This node-script should work as a substitution for `npm run` on bundles installation.

It is tested on Windows git-bash, powershell and cmd.exe.

Powershell and cmd.exe on windows need `node bundle-install` to run, which is now documented.

This is completly based on:
https://stackoverflow.com/questions/31773546/the-best-way-to-run-npm-install-for-nested-folders


seanohue:
I suggest having the `bundle-install` npm script execute this file, so it will make that script cross-platform compatible, rather than removing the NPM script.

shawncplus:
Tested on linux. Could someone with a mac/windows verify please

nahilep:
Should resolve #321 except the version bump on motd

PR #327, opened 2018-04-08
Fixed mapped rooms exits not working as a move command

### LINKED ISSUE #323: "up" and "down" not defined in coordinate rooms
### Description

While in coordinate based rooms, the full text of up and down fail when attempting to navigate

### Expected outcome

up should navigate up

down should navigate down

### Actual outcome

up returns "Huh?"

''

### Version of Ranvier

2.0 It hink

nahilep:
This is an attempt to fix #323

### DIFF
--- a/src/Player.js
+++ b/src/Player.js
@@ -152,21 +152,32 @@ class Player extends Character {
       return false;
     }
 
-    const exits = Array.from(this.room.exits).filter(e => e.direction.indexOf(direction) === 0);
-
-    if (!exits.length) {
-      return false;
-    }
-
-    if (exits.length > 1) {
-      return false;
-    }
-
     if (this.isInCombat()) {
       return false;
     }
 
-    return true;
+    if (Array.from(this.room.exits).filter(exit => exit.direction === direction).length === 1) {
+      return true;
+    }
+
+    const coords = this.room.coordinates;
+    if (coords) {
+      const directions = {
+        north: [0, 1, 0],
+        south: [0, -1, 0],
+        east: [1, 0, 0],
+        west: [-1, 0, 0],
+        up: [0, 0, 1],
+        down: [0, 0, -1],
+      };
+
+      const diff = directions[direction];
+      if (diff) {
+        return this.room.area.getRoomAtCoordinates(coords.x + diff[0], coords.y + diff[1], coords.z + diff[2]);
+      }
+    }
+
+    return false;
   }
 
   save(callback) {


shawncplus:
I'm going to hold off on this for a bit because, while I think this will ameliorate the issue, I think there may be a better overall solution. That solution possibly including moving this `Player.canGo` method out of core. Right now this method makes the core really opinionated about what directions the game can have so I need to think about it some more

nahilep:
That's cool.

I was thinking the same thing and would love to discuss ideas. What I saw was, that there was a check for `inCombat()`. I also thought, that it was too opinionated, but decided to leave it in there so the function is still doing what it did.

Turns out that  after `Player.canGo` the move command is triggered which in turn checks the same thing again.

shawncplus:
My current thought is to move it into the `ranvier-lib/` bundle, I'll think a bit more about it after work

nahilep:
I had a little time to examine code and think about this issue more.
Here my thoughts:
The command parser makes already a lot of assumptions and it should be moved to ranvier-lib.
There are two checks in the command parser for moving:
* Is the command a shortcut for a direction (`n` for `north`, `ne` for `northeast`)?
* Is the command an exit. This is done in `Player.canGo()` atm

The second logic could be moved to room and implement a `getExits()` there which merges the defined exits and the exits implied by a room with coordinates. All other places which use that directions object used to determine the coordinates of an adjacent room can be easily refactored I guess. Maybe we need some more helpers like `Room.hasExit('north')` or so for rooms with coordinates, but since automatically build rooms are part of the core, I think this can happen in the core also.

What do you think?

shawncplus:
* I agree about moving `CommandParser`. Moving it to `ranvier-lib/` is easy but tedious.
* I agree that `Room.getExits` should be a thing but I'm not certain on the format yet; perhaps something like:

```js
room.getExits()

[
  { roomId: "limbo:2", direction: "climb", inferred: false }, // hard coded/named direction
  { roomId: "mapped:hallway-north-1", direction: "northeast", inferred: true }, // coordinate generated exit
]
```

This would make `Room.getExits` know about compass directions for coordinates but agnostic about "named" directions like "climb" or "tapestry". My hope is that the `inferred` property would make it easier to detect which is which so if you have a named exist called "northeast" your bundle command could differentiate

shawncplus:
Closed as new `Room#getExits()` API is implemented in core

PR #333, opened 2018-04-13
Updated required NodeJS version

Vindexus:
I got this error when trying to install and run: `Error: Ranvier's core engine requires Node version >= 8.9.4`


shawncplus:
Awesome, much appreciated

PR #334, opened 2018-04-13
`commands` list sorted alphabetically

Vindexus:


shawncplus:
This is a bit nitpicky but could you do the same for the channel list while you're in there?

shawncplus:
:+1:

PR #337, opened 2018-04-14
Added auto login as a config option

Vindexus:


shawncplus:
This is fantastic. Could you update the server config docs at `docs/server_config.md`

Vindexus:
I actually redid this as a bundle you can see here: https://GitHub.com/Vindexus/ranvier-auto-login

I like its organization better. Still working on the installation, but it could just be copy and pasted in.

shawncplus:
I worry about having two `input-events` registering an `intro` listener could run into a race condition when emitting the `data` event, have you run into that issue?

Vindexus:
I have not yet. Would all the intro listeners be dealt with before any of
the emits they send out are actually broadcast? If not it could be an
issue.

shawncplus:
Yeah my suggested would be to keep what you had before since just by the nature of how events are registered I would consider the order in which the listeners are called is, while not exactly undefined behavior, close enough to undefined behavior. So any two event listeners really shouldn't expect another to be done or not

seanohue:
Not to throw in some scope creep, and I'll put up a separate issue for this, but:

It would be cool to be able to have players configure this individually as well.

For example, logging into a character and doing 

`config set autologin on` or something, would set that character to their default and skip the character selection until they turn that setting off.

seanohue:
@Vindexus regarding your question on event listeners:
`The EventEmitter calls all listeners synchronously in the order in which they were registered. This is important to ensure the proper sequencing of events and to avoid race conditions or logic errors. `
According to the Node.js docs.

So one listener would be called before the other, and would run and emit synchronously, and then the other would run and emit.

It is not clear when the listeners would fire that are emitted from the original listeners, but I think they would be emitted immediately and synchronously _unless_ they use process.nextTick, setImmediate, or similar APIs.

shawncplus:
Yeah the reason I say it's not exactly undefined behavior but close enough is because there's no guarantee of which order the `BundleManager` will load the input event listeners. _Technically_ they're ran in order but that order is not guaranteed

Vindexus:
Added documentation to server config. Let me know if I should expand it at all.

Vindexus:
Does this need any changes?

shawncplus:
There's nothing wrong with this per se except that that I want to rethink how `Config` and `ranvier.json` work in general and bundles are going to be moving to their own repos soon. If someone wants it later they can resubmit the PR to that repo

PR #338, opened 2018-04-15
Bundles can be loaded from node_modules

Vindexus:
Bundles can be installed like so `node install-bundle.js Vindexus/ranvier-final-strike`

### DIFF
--- a/install-bundle.js
+++ b/install-bundle.js
@@ -1,0 +1,79 @@
+#!/usr/bin/env node
+'use strict';
+
+const fs = require('fs');
+const resolve = require('path').resolve;
+const join = require('path').join;
+const cp = require('child_process');
+const os = require('os');
+
+const pkgPath = resolve(__dirname, 'package.json');
+const ranvierConfigPath = resolve(__dirname, 'ranvier.json');
+const ranvierConfig = require(ranvierConfigPath);
+let bundlePackage = process.argv[2].trim();
+
+function exit (msg) {
+  console.log(msg)
+  process.exit()
+}
+
+function addToRanvier() {
+  const packageDotJSON = require(pkgPath);
+
+  let pkgName;
+  const dependencies = new Map(Object.entries(packageDotJSON.dependencies))
+  console.log('bundlePackage',bundlePackage);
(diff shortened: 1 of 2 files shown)


shawncplus:
I definitely like this idea but I want to take some time to think about it. The reason is that right now, without this change, a bundle dir can have a `package.json` which is good, keeps the main `package.json` clean. But with this change it opens up the idea of having a ranvier bundle depend on another ranvier bundle inside its `package.json`, AKA meta-bundles. I'm not sure if this code would support that since it seems to only be looking at the first level inside `node_modules/`.

Also just as a pedantic note the command should probably be called something besides `install-bundle` given that there is already a `bundle-install`. Perhaps like `npm` itself there is just one `bundle-install` command and if you don't pass it any args it just does what it does now but if you do pass it a bundle to install it'll install it.

Vindexus:
>I'm not sure if this code would support that since it seems to only be looking at the first level inside node_modules

You're right I don't think it would, you'd need to install every dependency of the bundle you're installing.

Maybe the `setup-bundles.js` file could be extended to look for dependencies of any bundles you've installed with npm and add those to ranvier.json.

There's definitely some things to consider. I think if maybe we got some requirements written down I could work out a solution.

shawncplus:
My "requirements" would be: Easy to install, easy to remove, easy to fork

There are 3 ways one could possibly install a bundle

1) Just copy the files over. Pros: easy install/removal, easy fork (just copy the folder). Cons: Hard to receive updates to the bundle if there is a bugfix/feature
2) Git submodule. Pros: Easy to receive updates from upstream. Cons: Hard to edit locally, hard to manage
3) Node module. Pros: Easy to receive updates from upstream, easy install/removal. Cons: Impossible to edit locally without forking the original repo and making npm know about your version (either by publishing or by also distributing the other repo with your project)

At the moment the cons of node modules outweigh the pros in my opinion. I think for the time being if people want the easy upstream benefit from using a node module they can just use a git submodule in the bundles directory. Thoughts?

shawncplus:
Closing this as we've settled on git submodules along with accompanying `install-bundle`, `remove-bundle` scripts

PR #343, opened 2018-04-17
WIP: Move CommandParser out of core

nahilep:
This is not final, I need to discuss something first.

I have `const CommandType = require('../bundles/ranvier-lib/lib/CommandType');` in src/Command.js to make it run. Obviously this is a no-go. How would I approach this the best way?

Also removed unused requires in files touched.


shawncplus:
The only suggestion I would have right now is to keep `CommandType` in the core, I'll have to think about it some more if it's possible to move that out or if commands should even have a type. The way the bundles use `CommandType` is a bit strange

nahilep:
From what I can see `CommandType` is only used in `CommandParser` and in `src/Command.js` to set a default value for `command.type` in `constructor()` -> `this.type = def.type || CommandType.COMMAND;`

If we could get rid of this default assignment we wouldn't need it in `src/`.

shawncplus:
That's definitely the case but I worry that would break things if we just removed the default. I'll have to look into it more after work

shawncplus:
Closing this as this has been implemented in core

Issue #329, opened 2018-04-09
Fix Abilities docs


shawncplus:
The abilities docs page currently calls `player.calculateWeaponDamage()` which doesn't exist anymore. Doc should be updated to reflect the actual `bundles/ranvier-classes/skills/lunge.js`

shawncplus:
Closed with #346

Issue #324, opened 2018-03-29
Intro screen displays incorrect current version for codebase


MarchosiasM:
The login screen displays ..1.2, or so, I believe?

shawncplus:
That's located in `data/motd` this is a great issue to test the pull request waters for someone new

MarchosiasM:
Heard

shawncplus:
Closed with #346

PR #346, opened 2018-04-19
Issues #329 and #324

sezycei:


shawncplus:
lunge.js doc needs to include the `const Combat = require('../../ranvier-combat/lib/Combat')` line line the lunge skill does. Also for the "this function calculates damage" comment I suggest something more helpful like "An example damage calculation using the Combat library from the `ranvier-combat` bundle"

Issue #342, opened 2018-04-17
Move "critical" code from core Damage class


shawncplus:
The "critical" code inside `Damage` is very opinionated in that it requires a "critical" attribute and assumes that the attribute is a percentage chance.

All of this could be handled at the time of instantiating the `Damage` instance and just configuring the initial damage amount instead of waiting for `evaluate`. As such all this logic can be moved into the `ranvier-combat` bundle

PR #347, opened 2018-04-19
Moved critical hit logic from Core to Combat Bundle

sezycei:
Issue #342 resolved by moving critical hit logic from src/Damage.js to bundle/ranvier-combat/lib/Combat.js

I tested it a few times on fresh installs before making the pull request. If you see any issues with it, please make sure to let me know. This uses most of the same functions and maintains the exact same end-user functionality of the combat and critical system as before the modification.

### DIFF
--- a/bundles/ranvier-combat/lib/Combat.js
+++ b/bundles/ranvier-combat/lib/Combat.js
@@ -92,14 +92,14 @@ class Combat {
    * @param {Character} target
    */
   static makeAttack(attacker, target) {
-    //defines variables associated with critical strikes based on a random percentage chance.
+    // defines variables associated with critical strikes based on a random percentage chance.
     const critChance = Math.max(attacker.getMaxAttribute('critical') || 0, 0);
     const critical = Random.probability(critChance);
     const critMultiplier = critical ? 1.5 : 1;
     const amount = Math.ceil(this.calculateWeaponDamage(attacker) * critMultiplier);
     const damage = new Damage({ attribute: 'health', amount, attacker });
     
-    //implements critical strike if rolled successfully
+    // implements critical strike if rolled successfully
     if (critical) {
       damage.critical = true;
     }
--- a/bundles/ranvier-combat/lib/Combat.js
+++ b/bundles/ranvier-combat/lib/Combat.js
@@ -92,28 +92,26 @@ class Combat {
    * @param {Character} target
    */
   static makeAttack(attacker, target) {
-    const amount = this.calculateWeaponDamage(attacker);
-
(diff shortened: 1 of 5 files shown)


bundles/ranvier-combat/lib/Combat.js
```diff
@@ -91,8 +92,19 @@ class Combat {
...
-    const amount = this.calculateWeaponDamage(attacker);
-    const damage = new Damage({ attribute: 'health', amount, attacker });
+    var amount = this.calculateWeaponDamage(attacker);
+
+	//defines variables associated with critical strikes based on a random percentage chance.
+    const critChance = Math.max(attacker.getMaxAttribute('critical') || 0, 0);
+    const critMultiplier = 1.5;
+    const critical = Random.probability(critChance);
+    
+	const damage = new Damage({ attribute: 'health', amount, attacker });
```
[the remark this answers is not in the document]
shawncplus: Also, you should change the `amount` _before_ instantiating `new Damage()` instead of instantiating it with the pre-critical amount then modifying it after the fact

shawncplus:
The comments and `if (critical)` block still have wrong indentation lol

docs/extending/classes.md
```diff
@@ -73,6 +75,11 @@ module.exports = (srcPath) => {
   // change them if you need to.
   const damagePercent = 250;
   const energyCost = 20;
+  
+  //An example damage calculation using the Combat library from the ranvier-combat bundle
```
seanohue: for the sake of consistency in documentation, please place a space after the `//` in comments (this is consistent with airBnB and standardJS styles also)

bundles/ranvier-combat/lib/Combat.js
```diff
@@ -91,18 +92,26 @@ class Combat {
    * @param {Character} target
    */
   static makeAttack(attacker, target) {
-    const amount = this.calculateWeaponDamage(attacker);
+    //defines variables associated with critical strikes based on a random percentage chance.
+    const critChance = Math.max(attacker.getMaxAttribute('critical') || 0, 0);
```
seanohue: Pulling each of these calculations into its own constant makes this way more readable than it would be otherwise, nicely done.

sezycei: I'm not sure if you're asking me to change something or are just complimenting how I did this.

seanohue: I thought about suggesting pulling all crit-related code into a function like `this.calculateCritical(attacker, target, amount)` but overall just thought it was well done.

seanohue:
I'm wondering why this was closed?

PR #349, opened 2018-04-20
Docs improvement for node install

### LINKED ISSUE #308: Even more stupid-proofing for try it out quickly on Ubuntu
### Description

Literally following the "try it out quickly" steps has two problems:

1.

npm install -> the stupid user did not actually have node.js installed; this one is easy even for stupid people as ubuntu guesses what they did wrong and prompts them to run:

sudo apt install npm

2.

./ranvier
/usr/bin/env: node: No such file or directory

The stupid user did not expect npm install on ubuntu not to include the "node" command but it actually doesn't, because the "node" command is already claimed by an old Amateur Packet Radio Node program. (No, seriously.)

The user can mitigate this by either:

A. editing ranvier and changing "node" to "nodejs" in line 1
B. sudo apt install nodejs-legacy

### Expected outcome

Even a stupid user can get "try it out quickly" to run.

### Actual outcome

The stupid user does not quite understand why ./ranvier gives the error about "node" or immediately how to fix it.

### Steps to reproduce

On a shiny new Ubuntu server try to follow "try it out quickly". See that npm install doesn't work (until the stupid user installs npm). Then see that ./ranvier does not work as "node" is not a valid command in their path.

### Version of Ranvier

 ./ranvier --version
/home/ubuntu/ranviermud/src/Account.js:129
    const {
          ^

SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:374:25)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/home/ubuntu/ranviermud/src/AccountManager.js:4:17)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)

seanohue:
Resolves #308 , fixes a thing about commas, and adds a reference to guide potential contributors.

### DIFF
--- a/docs/get_started.md
+++ b/docs/get_started.md
@@ -12,7 +12,7 @@ npm install
 npm run bundle-install
 ```
 
-If you would like to use the latest, but perhaps not stable, feature,s make sure you are on the
+If you would like to use the latest, but perhaps not stable, features: make sure you are on the
 `staging` branch.
 
 It is also recommended that contributors submit PRs from the `staging` branch. For more on contributing, see [CONTRIBUTING.md in the Ranvier Github repo](https://github.com/shawncplus/ranviermud/blob/staging/CONTRIBUTING.md).
--- a/docs/get_started.md
+++ b/docs/get_started.md
@@ -2,6 +2,8 @@
 
 ## Installation
 
+Please make sure to have the [Node LTS](https://nodejs.org/en/) installed.
+
 ```sh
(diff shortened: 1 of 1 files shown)


shawncplus:
Closing, going to require 10.x going forward

PR #350, opened 2018-04-20
Startquest behavior

seanohue:
Figured I'd give back to the community by PRing this handy quest behavior I whipped up for the MUD jam...


shawncplus:
This is pretty old and conflicts now, also the doc changes made it seem like the behavior was built into the core. Closing for now

PR #352, opened 2018-05-02
Make scan work for coordinate-based rooms.

seanohue:
This ended up with some extra commits that have already been merged, due to lazy PR process on my part I suppose.

Anyway, prior to this using `scan` in coordinate-based rooms did nothing, now it is implemented for all sorts of rooms. This also provides another example implementation of how to do coordinate-based stuff.


shawncplus:
I'm willing to bet this works despite it sitting open for a while but that coordinates to direction matrix is in like 5 places now and I hate it

PR #372, opened 2018-11-17
Fix "npm run start" for windows users

nahilep:
Even though I can run just `./ranvier` because I have node in my PATH `npm run start` fails.
Prepending `node` to the npm script fixes this for windows and shouldn't harm other OS

Please let me know if you want this onto staging and not on npmified

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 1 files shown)


shawncplus:
I intentionally removed jsdoc because that code won't exist in this repo going forward, with that said that means you can also remove the build-docs from the package.json

PR #382, opened 2019-01-12
Fixed issue with Wally the vendor

markscho:
Wally the vendor would not list his wares properly using the LIST command.  This is because one of his items was improperly referenced (a typo).


shawncplus:
I feel like I reject everyone's first PR against the project, it's not personal. 

This branch is effectively frozen, so this bug fix should instead be applied to the 3.x version of `ranvier-areas` which is https://github.com/RanvierMUD/bundle-example-areas repo. Unfortunately I've already made that fix there :/

PR #383, opened 2019-01-20
Add util script to run `npm install` in all bundles

seanohue:
This is good for when (for any reason), your npm dependencies get blown up in the submodules.


digitalsadhu:
Should there be a matching run script for this in package.json?

seanohue:
Good idea @digitalsadhu -- added it.

shawncplus:
I don't think this is necessary since you can just run `git submodule foreach npm install`

shawncplus:
@seanohue I had to update package.json to fix the missing preinstall for `link-module-alias` so you'll have to update your package.json.

seanohue:
TIL that `git submodule foreach` is a thing. Also, the missing preinstall was probably the root cause of the issue that made me have to create this command. I may just change the `npm` script to do `git submodule foreach npm install` and then get rid of the JS file.

PR #388, opened 2019-03-28
Update .nvmrc

Avienneau:
The project says it needs Node v10.12.0 to work. With a fresh install and typing `nvm use`, I get set to Node 7, which causes `npm run init` to fail.


shawncplus:
Forgot this file even existed, thanks for the update!

PR #397, opened 2020-05-21
Rolling rounds

MarchosiasM:
Well this is going to be a nightmare to review so I'm not going to, and then just roll with that.


MarchosiasM:
Woops, once again meant for my own repository

PR #413, opened 2026-02-05
Starting Mud

rendall:
This is the minimum to get a working telnet server started, that can be logged into and receive a response.


rendall:
*Sigh*  Sorry.  This was (another) error.  GitHub adds *this* repo as the merging destination by default, and not the main branch of my own fork.  I'll be more careful.

PR #414, opened 2026-02-06
Fix install script

rendall:
A post-install script was accidentally introduced in a previous commit.  This removes it.


rendall:
G'dammit.  I really don't know how forks work.  

In the meantime, these updates are meant for https://github.com/Ranvire/ranviermud

My apologies again.

